Issue #7, opened 2018-09-21
Authentication, User Roles and Login


banukutlu:
In GitLab by @bzk60 on Jul 17, 2018, 17:21

https://github.com/psu-libraries/psu_dir?  
https://github.com/psu-libraries/devise_remote?

Required by psu-libraries/psulib_blacklight#6

ruthtillman:
Works for me from my login.

Issue #11, opened 2019-10-31
README


cdmo:
document webservice version, ruby and rails versions, etc. 

how to start the application

PR #18, opened 2019-11-12
authentication

banukutlu:
part of issue #7 
also needs specs #19


cdmo:
yes, needs specs: https://github.com/psu-libraries/myaccount/issues/19

cdmo:
What do you think about updating the README.md with some descriptions of some of this stuff, particularly that we are using Warden and how it works with `config` and what you have to do to make that work locally? Other stuff?

Gemfile
```diff
@@ -6,10 +6,13 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '2.6.5'
 
 gem 'bootsnap', '>= 1.4.2', require: false
+gem 'config'
+gem 'http'
```
cdmo:
```suggestion
gem 'http', '~> 4.0'
```

Let's peg to a version. That ok?

cdmo: Same conversation for `warden` and `config` too of course

cdmo:
What are the changes to the yml files? Like `config/settings/development.yml`? Just empty files... 🤔

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class SessionsController < ApplicationController
+  def index
+    # @symphony_ok = symphony_client.ping
```
cdmo:
```suggestion
```

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    redirect_to summaries_url if current_user? || request.env['warden'].authenticate(:library_id)
+    # TODO: redirect_to errors_url, alert: 'Unable to authenticate.'
```
cdmo: see #20

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+    # TODO: flash.now.alert = request.env['warden'].message
```
cdmo: What is there todo for this one?

banukutlu:
> What are the changes to the yml files? Like `config/settings/development.yml`? Just empty files... 🤔

yes they are empty so far, got added as part of config:install

app/controllers/sessions_controller.rb
banukutlu: for the case if login was not successful, display the message kinda thing 🤔

cdmo: So we can delete, because I created #20 ?

banukutlu: sure, will do

banukutlu:
> Requesting a README update (I guess this is part of #11)

updated 👍

PR #23, opened 2019-11-18
Adds initial Travis configuration

cdmo:
#22


.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+language: ruby
+rvm:
+  - 2.6.5
+bundler_args: --without development
+sudo: false
+cache:
+  bundler: true
+  yarn: true
+  directories:
+    - tmp/solr_dl/
```
banukutlu: @cdmo we do not need this line, right?

cdmo: Right! Thanks for catching.

Issue #19, opened 2019-11-13
Auth specs


cdmo:
specs for methods in #18 

- [x] app/controllers/application_controller.rb
- [x] app/controllers/sessions_controller.rb
- [x] app/controllers/summaries_controller.rb
- [x] app/models/patron.rb
- [x] app/models/user.rb
- [x] app/services/symphony_client.rb

Issue #1, opened 2019-10-28
Account Overview Page


ruthtillman:
Page: Overview 

Display on login/entering My Account

* [x] Messages: problems with account
  * [x] Patron Standing Alerts for `DELINQUENT`, `BARRED`, `BLOCKED` and `COLLECTION`
* [x] Fines (total owed in $)
* [x] Total items checked out
   * [x] Total items Overdue
* [x] Total items on hold
  * [x] Total item ready for pickup
* [x] Additional links

banukutlu:
What are the possible Alert messages for patron standing?
and for which ones there should be an alert message on the summary page?

ruthtillman:
@banukutlu that is a good question. Waiting to see if @mkutch has any other answers, but I have only seen Delinquent on my account.

The two real negative statuses are BARRED and BLOCKED. I believe BLOCKED corresponds to DELINQUENT. I believe BARRED means they need to opt into the garnishment -- is this correct, Maryam? So in that case, it we might need to put in a link for opting into garnishment.

banukutlu:
@mkutch what are the other alerts (other than delinquent) that are displayed on the summary page currently?

mkutch:
Confirmed with Chris that DELINQUENT, BARRED and BLOCKED are the onyl alerts. There are certain thresholds which the library has set that triggers which status a patron will be. We are testing something else  to see if another condition/alert shows up . Will let you know.

mkutch:
When a user is sent to COLLECTION this is the alert.

![image](https://user-images.githubusercontent.com/43478453/70470879-063a6f00-1a9a-11ea-9d5f-ebbbbef82b42.png)

ruthtillman:
Seems good.

PR #29, opened 2019-11-20
initial account overview work

banukutlu:
Part of Issue #1 

## Additional change

* Added model specs for user and patron as part of #19

### DIFF
--- a/app/views/summaries/_summary.html.erb
+++ b/app/views/summaries/_summary.html.erb
@@ -1,4 +1,18 @@
 <div class="page-section">
+  <h3>Checkouts: <%= patron.checkouts.length %></h3>
+
+  <% count = patron.checkouts.select(&:recalled?).length %>
+  <% if count.positive? %>
+      <div class="ml-4"><%= count %> recalled</div>
+  <% end %>
+
+  <% count = patron.checkouts.select(&:overdue?).length %>
+  <% if count.positive? %>
+      <div class="ml-4"><%= count %> overdue</div>
+  <% end %>
+</div>
+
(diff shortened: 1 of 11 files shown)


cdmo:
Actually, never mind, I see that the user and patron specs are bound in with some of the other work here. I'm making my way through.

spec/models/patron_spec.rb
```diff
@@ -3,5 +3,64 @@
 require 'rails_helper'
 
 RSpec.describe Patron, type: :model do
```
cdmo:
```suggestion
RSpec.describe Patron do
```
because of `config.infer_spec_type_from_file_location!`

cdmo: Fixed in #32

cdmo:
Just to capture some of our conversation on Slack: there is probably some extraction of `Hold` and `Checkout` we can do to either `BibRecord` or perhaps a new model called `SirsiTransaction` (or something like that). Consider for a refactor later.

app/views/summaries/_summary.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="page-section">
+  <h3>Checkouts: <%= patron.checkouts.length %></h3>
+
+  <% count = patron.checkouts.select(&:recalled?).length %>
```
cdmo: I wonder why this is in a view and not the controller (or a helper) 🤔

banukutlu: hmm yes let's move it when we are working on the views

PR #33, opened 2019-11-25
Moar sws specs

cdmo:
relates to #19

### DIFF
--- a/spec/services/symphony_client_spec.rb
+++ b/spec/services/symphony_client_spec.rb
@@ -10,47 +10,54 @@ RSpec.describe SymphonyClient do
       stub_request(:post, Settings.symws.url + '/user/patron/login')
         .with(body: Settings.symws.login_params.to_h,
               headers: Settings.symws.headers)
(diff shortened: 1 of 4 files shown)


spec/services/symphony_client_spec.rb
```diff
@@ -7,50 +7,57 @@
...
+
+    context 'when requesting item details' do
+      it 'requests the item details for checkouts' do
+        client.patron_info(user, item_details: { circRecordList: true })
+
+        expect(WebMock).to have_requested(:get, "#{Settings.symws.url}/user/patron/key/some_patron_key")
+          .with(query: hash_including(includeFields: match(/circRecordList{.*,item{.*}}/)))
+      end
+
+      it 'requests the item details for requests' do
```
banukutlu:
```suggestion
      it 'requests the item details for holds' do
```

cdmo: 🤔

cdmo: Yeah that's better 👍 thanks!

Issue #24, opened 2019-11-19
 Renew session functionality


cdmo:
potential approach: Ping symphony to see if the user token expired or not

cdmo:
@banukutlu please tell me we just forgot to close this before right? RIGHT? 😆

banukutlu:
> @banukutlu we just forgot to close this before right?

I think there was still some things we wanted to clear out but I honestly think it is fine, guess we both do not wanna change/refactor right now :) So yeah let's have this close :)

PR #35, opened 2019-11-26
Renew session

cdmo:
#24 

I left the `#logout` on `SessionsController`. I figured we may actually want to keep it around either because we will present this to our users as an option for logging out OR for our own local dev purposes? Not sure.

Also, please do look for another way of doing this! I'm all ears.

Lastly, in order to simulate an invalid session you can throw a byebug/pry on `ApplicationController#current_user` and set a bad patronKey like 

`[1] pry(#<SessionsController>)> session_data['session_token'] = '83838'`

And that will invalidate the session.

## Additional change

Adds a comment on session/user management.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  helper_method :current_user, :current_user?, :patron, :symphony_client
+  helper_method :current_user, :current_user?, :patron, :renew_session_token, :symphony_client
 
   def current_user
     session_data = request.env['warden'].user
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  helper_method :current_user, :current_user?, :patron, :symphony_client
+  helper_method :current_user, :current_user?, :patron, :renew_session_token, :symphony_client
 
   def current_user
     session_data = request.env['warden'].user
+    # Assuming the && is used solely to guard against nil. A new User is minted with every request based upon details
+    # Warden set in the session data in cookies. So, new user every request, same SWS session data.
```
banukutlu: :100: :D

PR #36, opened 2019-12-03
Renew session refactor

banukutlu:
@cdmo wondering what you would think about these changes? 
related to #24

### DIFF
--- a/spec/models/patron_spec.rb
+++ b/spec/models/patron_spec.rb
@@ -3,13 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe Patron do
-  subject(:patron) do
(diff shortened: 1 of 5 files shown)


spec/models/patron_spec.rb
```diff
@@ -40,6 +40,10 @@
     expect(patron.barcode).to eq '1234'
   end
 
+  it 'has a valid session' do
+    expect(patron).not_to be_stale
```
cdmo: niiice

banukutlu: rubocop recommended ;)

cdmo: 😅

PR #37, opened 2019-12-04
account overview alerts

banukutlu:
part of Issue #1 - adding patron alerts to summary page

### DIFF
--- a/spec/views/summaries/index.html.erb_spec.rb
+++ b/spec/views/summaries/index.html.erb_spec.rb
@@ -19,6 +19,7 @@ RSpec.describe 'summaries/index', type: :view do
       protected
 
         def patron; end
+
         def current_user?; end
 
         helper_method :patron, :current_user?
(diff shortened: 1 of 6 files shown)


spec/views/summaries/index.html.erb_spec.rb
```diff
@@ -3,5 +3,49 @@
...
+  before do
+    controller.singleton_class.class_eval do
+      protected
+
+        def patron; end
+
+        def current_user?; end
+
+        helper_method :patron, :current_user?
+    end
```
cdmo: What's all this then?

cdmo: ☝️ was meant to be a joke, but I think I'll need you to walk me through this spec

banukutlu: sure, we can talk about but briefly this is to access to the inherited (in this case from application controller) helper_method helpers from the view spec. 
See here: https://relishapp.com/rspec/rspec-rails/v/2-0/docs/view-specs/view-spec#spec-with-view-that-accesses-helper-method-helpers

SUL also uses this same method in their spec and I could not find another way to do it but there might be another better way to do it of course so pls let me know if you have an idea.

PR #39, opened 2019-12-04
Procfile for foreman

banukutlu:


banukutlu:
we decided that we do not need foreman for myaccount so closing this PR.

PR #40, opened 2019-12-05
Fines summary

cdmo:
#1 

Notes:
* didn't think we should reach into the Fines class during testing of Patron
* would like to refactor to extract "Sirsi System Items" into a class and take out things like key
* need to create view specs (what's in the View is demo purposes anyway)

Screenshot 2019-12-05 11 39 12

## Additional change

Removes mailers

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,6 +1,0 @@
-# frozen_string_literal: true
-
-class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
-  layout 'mailer'
-end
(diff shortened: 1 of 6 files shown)


cdmo:
Another note:

rubocop wants me to use `be` instead of `eql` but they aren't the same. `be` will fail (object identity) and `eql` doesn't (value identity). It is okay with `eq` which is value identity with coercion. Not sure what it has against value identity without coercion.

```rb
irb(main):004:0> require 'bigdecimal'
=> true
irb(main):006:0> a = BigDecimal(12) # this is what ruby does to the value from Sirsi for fee
=> 0.12e2
irb(main):010:0> a == 12 # this is the same as `eq` in rspec
=> true
irb(main):013:0> a.eql? 12
=> true
irb(main):016:0> a.equal?12 # this is `be` in rspec 
=> false
irb(main):017:0> 12.class # just to show
=> Integer
```

PR #46, opened 2019-12-10
4 fines

cdmo:
Some stuff about FactoryBot. I think the main reason to use it, beyond potentially speeding up the test suite, would be to eliminate the "[Mystery Guest](https://thoughtbot.com/blog/mystery-guest)" problem.

I watched a couple tutorials to help me get going with this gem:

* https://thoughtbot.com/upcase/videos/factory-bot
* https://www.youtube.com/watch?v=x2qapy3s6Tg

But, the best post I read on this was "[ Tips for Using FactoryGirl Without an ORM ](https://thoughtbot.com/blog/tips-for-using-factory-girl-without-an-orm)" (exactly our situation :))

So, I like it in the end. I have now had some thoughts about refactoring too. Like, why not set up values like `owed_amount` as an attribute instead of a special method. Would make mocking even easier/clearer. I could just set `owed_amount: "12.00"` because the factory would have a special attribute for that. 

## Additional Change

Fix renewal of session problem (see d243c7d )

## one more thing

Feel like view specs have to ignore describe class rule from rubo... 🤷‍♂️


spec/models/fine_spec.rb
```diff
@@ -30,7 +30,19 @@
     }
   end
 
-  it 'has a fee' do
-    expect(fine.fee).to eq 12
+  it 'has tallies all the fees into a total amount owed' do
```
banukutlu:
```suggestion
  it 'tallies all the fees into a total amount owed' do
```

spec/views/fines/index.html.erb_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    it 'displays fine\'s reason correctly' do
+      fine.record['fields']['block']['key'] = 'DAMAGE'
+      render
+      expect(rendered).to include 'Fee for damaged materials'
+    end
+
+    it 'displays fine title' do
+      fine.record['fields']['item']['fields']['bib']['fields']['title'] = 'A wonderful title'
```
banukutlu: wonderful 😆

app/views/fines/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <th scope="col">Reason</th>
+    <th scope="col">You owe</th>
+    <th scope="col">Date billed</th>
+  </tr>
+  </thead>
+
+  <tbody>
+    <% @fines.each do |fine| %>
+    <tr>
+      <th scope="row"><%= fine.title %></th>
```
banukutlu: I assume title will include link to record, might need to confirm with Ruth maybe 🤔

app/controllers/fines_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class FinesController < ApplicationController
+  before_action :authenticate_user!
+
+  # Render a patron or groups fines or checkouts
```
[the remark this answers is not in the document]
cdmo: oops. nice catch

spec/models/fine_spec.rb
cdmo: nice.

spec/views/fines/index.html.erb_spec.rb
cdmo: I am so creative!

app/views/fines/index.html.erb
cdmo: Oh good catch, no, let's just fix it. I'll add it in.

cdmo: Looks like item type is under title too. My mistake.

cdmo: fixed in latest

app/services/symphony_client.rb
```diff
@@ -33,15 +33,21 @@ def patron_info(user, item_details: {})
...
+          'blockList{*}',
+          "holdRecordList{*,#{ITEM_RESOURCES if item_details[:holdRecordList]}}",
+          'circRecordList{*,circulationRule{loanPeriod{periodType{key}},renewFromPeriod},' \
+          "#{ITEM_RESOURCES if item_details[:circRecordList]}}"
+        ]
+      end
     end
 
     def authenticated_request(path, headers: {}, **other)
+      # binding.pry
```
banukutlu:
```suggestion
```

cdmo: dangit

cdmo:
I need to rebase

cdmo:
Wow a gold medal! Thanks Banu! 😆

Issue #6, opened 2019-10-28
Checkouts/Renewals


ruthtillman:
Page: Checkouts/Renewals

* [x] View list of items checked out
* [x] Renew individual (check box)
* [x] Renew all - #74 
* [x] Custom error messages - #69 

Info on what we currently have displayed here: https://github.com/psu-libraries/psulib_blacklight/wiki/My-Account-Components#checkoutsrenewals

banukutlu:
@ruthtillman @mkutch What are the possible statuses for the checkouts table other than `Overdue`? 

![Screen Shot 2019-12-09 at 11 58 37 AM](https://user-images.githubusercontent.com/6477867/70456059-6837ac00-1a7b-11ea-91b3-7a9e31a06e11.png)

@mkutch How are the statuses determined currently? There is Active/Inactive status field but also Overdue field?

ruthtillman:
Only other one I can think of is Recalled. but maybe there's a list which has values? @mkutch ?

mkutch:
talked with Chris and the other status besides "Overdue" that was recently implemented is "Claims returned" ... There should be one on that account, but isn't being displayed in my library account (separate issue I need to look into that)

banukutlu:
@mkutch another question re `You owe` column on the checkouts page. Could you pls find a test account that has accrued? And could you pls confirm if I use the `estimateOverdueAmount ` field for this column with no other calculation?

mkutch:
Yes that is the estimateOverdueAmount. I put an example in box for you.

mkutch:
The logic regarding Status of items.  If `overdue` is `true` then check if `claimsReturnedDate` is NOT null then the status is `Claims returned` else status is `Overdue`.

ruthtillman:
So are recalls just changed due dates or is there anything special about them?

mkutch:
Recalled item will have a new due date and looks like this

![image](https://user-images.githubusercontent.com/43478453/70534550-4d237580-1b29-11ea-9e41-90465e0f824c.png)

Or if Overdue AND Recalled like this

![image](https://user-images.githubusercontent.com/43478453/70534590-60cedc00-1b29-11ea-9613-789216249d68.png)

ruthtillman:
Thank you @mkutch ! And is there something in the due date response that'll have the recalled stuff clearly?

mkutch:
Check if `recallDueDate` is NOT null.

banukutlu:
[...]

PR #47, opened 2019-12-11
checkouts

banukutlu:
Issue #6 - Checkouts (does not include renewals)


app/models/concerns/bib_record.rb
```diff
@@ -89,6 +89,10 @@ def item_key
     fields.dig('item', 'key')
   end
 
+  def item_type
+    item.dig('itemType', 'key')
+  end
+
```
cdmo:
```suggestion
```

This can be removed, see `item_type_code`

spec/controllers/checkouts_controller.spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      warden.set_user(user)
+      allow(mock_patron).to receive(:checkouts).and_return(checkouts)
+    end
+
+    it 'renders the index template' do
+      get(:index)
+      expect(response).to render_template 'index'
+    end
+
+    it 'assigns a list of fines' do
```
cdmo:
```suggestion
    it 'assigns a list of checkouts' do
```

spec/models/patron_spec.rb
```diff
@@ -49,7 +49,7 @@
 
   context 'with checkouts' do
     before do
-      fields[:circRecordList] = [{ key: 1, fields: {} }]
+      fields[:circRecordList] = [{ key: 1, fields: { status: 'ACTIVE' } }]
```
cdmo: I'm confused about this. The commit message says you were trying to fix active checkouts - but, no update to the specs themselves, just the stubbed item here... ?

app/models/checkout.rb
```diff
@@ -30,16 +34,28 @@ def recalled?
...
 
   def overdue?
     fields['overdue']
   end
 
-  def library
-    fields['library']['key']
+  def accrued
+    fields.dig('estimatedOverdueAmount', 'amount').to_d
+  end
```
cdmo: I'm curious about this because it is using a different method than `Fines` for determining owed amount... 🤔

app/services/symphony_client.rb
```diff
@@ -35,12 +35,16 @@ def patron_info(user, item_details: {})
     def patron_linked_resources_fields(item_details = {})
       case item_details
       when ->(h) { h[:blockList] }
-        ["blockList{*,#{ITEM_RESOURCES if item_details[:blockList]}}"]
+        ["blockList{*,#{ITEM_RESOURCES}}"]
+      when ->(h) { h[:circRecordList] }
+        ["circRecordList{*,#{ITEM_RESOURCES}}"]
+      when ->(h) { h[:holdRecordList] }
+        ["holdRecordList{*,#{ITEM_RESOURCES}}"]
```
cdmo: Good catch! Thanks.

app/models/checkout.rb
cdmo: For the main test account we use the Fines are different from the "You own" on checkouts 🤔

cdmo: resolved in Slack. the you owe in checkouts is like a "pending charges" really.

app/models/checkout.rb
```diff
@@ -22,6 +18,14 @@ def status
...
+    fields['dueDate'] && Time.zone.parse(fields['dueDate'])
+  end
+
+  def recall_due_date
+    fields['recallDueDate'] && Time.zone.parse(fields['recallDueDate'])
+  end
+
   def recalled_date
     Time.zone.parse(fields['recalledDate']) if fields['recalledDate']
   end
```
cdmo: Just want to point out that we aren't using this method anywhere except in `recalled?`. Seems like a good candidate to demote to `private`. Unless we want to display this information.

app/models/checkout.rb
```diff
@@ -30,16 +34,28 @@ def recalled?
     recalled_date.present?
   end
 
-  def patron_key
-    fields['patron']['key']
+  def claims_returned_date
+    Time.zone.parse(fields['claimsReturnedDate']) if fields['claimsReturnedDate']
+  end
```
cdmo: This could also potentially be `private` I think.

app/models/checkout.rb
```diff
@@ -30,16 +34,28 @@ def recalled?
...
+
+  def claims_returned?
+    claims_returned_date.present?
+  end
+
+  def renewal_date
+    Time.zone.parse(fields['renewalDate']) if fields['renewalDate']
   end
```
cdmo: Again, another potential `private` method since we don't display this. Private methods generally don't need to be tested either, or can be tested indirectly (by the method invoking it) which is why I bring it up mainly.

cdmo:
I know this isn't part of the PR, but can you remove https://github.com/psu-libraries/myaccount/pull/47/files?file-filters%5B%5D=.rb#diff-89af8c2c376b76346861341d2faaa0aaR158-R160 ? Shelf key isn't a method on the model, don't believe it's something we'll use.

app/models/concerns/bib_record.rb
banukutlu: oops, thought I got this deleted , thanks for catching it :)

spec/controllers/checkouts_controller.spec.rb
banukutlu: another oops :D

spec/models/patron_spec.rb
banukutlu: Which specs you expect to see changes? The only place I was able to easily filter out the inactive checkouts is in patron model checkouts method so I had to include status here.  Of course I might be missing sth. Maybe here there could be another spec for status Inactive to test if it is in the checkouts list or not but I kinda thought it would not matter much but I might be wrong.

cdmo:
I see that call_number, barcode and shelf_key all reside in bib_record and thus they are being tested here as bib_record is being included. However, I still sorta think we should get rid of all of them since we aren't using them at all 🤔

app/models/checkout.rb
banukutlu: It is used in the helper https://github.com/psu-libraries/myaccount/blob/cd4d4ac08f3ac59ac58aafd4f81e32902b87f6d9/app/helpers/checkouts_helper.rb#L21

app/models/checkout.rb
banukutlu: yes, this can be private 👍 thanks for catching it. I will update it and remove the related specs.

app/models/checkout.rb
cdmo: Ah, I see. Don't know how I missed that. Sorry.

banukutlu:
> I see that call_number, barcode and shelf_key all reside in bib_record and thus they are being tested here as bib_record is being included. However, I still sorta think we should get rid of all of them since we aren't using them at all 🤔

yes makes sense to get rid of those, I will clean those up 👍

app/models/checkout.rb
banukutlu: for this PR, renewal date actually is not required, I will remove it for this, and include it with renewals.

cdmo: Oh yeah, you're right. 👍

spec/helpers/checkouts_helper_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+require 'rails_helper'
+
+RSpec.describe CheckoutsHelper do
+  let(:checkout) do
+    instance_double(
+      Checkout,
+      title: 'Test Title',
+      catkey: 123,
+      due_date: Time.zone.parse('2019-11-14T23:59:00-05:00'),
+      overdue?: false,
```
cdmo: I literally didn't know you could do this. Awesome.

app/helpers/checkouts_helper.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Helper for checkouts views
+module CheckoutsHelper
+  def render_checkout_title(checkout)
+    title = checkout.title
+    title += " / #{checkout.author}" if checkout.author.present?
+    link_to title, "https://catalog.libraries.psu.edu/catalog/#{checkout.catkey}"
+  end
```
cdmo: I like this. I wonder if we should just use this in fines too just for consistency.

app/models/concerns/bib_record.rb
```diff
@@ -89,18 +77,6 @@ def item_key
...
-  end
-
-  def current_location
-    item.dig('currentLocation', 'key')
-  end
-
-  def lost?
-    current_location == 'LOST-ASSUM'
-  end
-
```
cdmo: oh cool, yeah these too, nice 😎

app/models/patron.rb
```diff
@@ -31,7 +31,9 @@ def display_name
   end
 
   def checkouts
-    @checkouts ||= fields['circRecordList'].map { |checkout| Checkout.new(checkout) }
+    @checkouts ||= fields['circRecordList']
+      .map { |checkout| Checkout.new(checkout) }
+      .select { |checkout| checkout.status == 'ACTIVE' }
```
cdmo: ah i see this detail now.

spec/models/patron_spec.rb
cdmo: I get it now. Thanks.

app/helpers/checkouts_helper.rb
banukutlu: I personally prefer the consistency as well like you. The current system has it like you had in fines with linking the word `Details`. I went on my own a bit here :p Let's see what Ruth will think about this way?

cdmo: Yep,I agree.

spec/views/fines/index.html.erb_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-RSpec.describe 'fines/index.html.erb' do
+RSpec.describe 'fines/index.html.erb', type: :view do
```
cdmo: danke

PR #48, opened 2019-12-16
Heroku experiment

cdmo:
Removes all traces of ActiveRecord and databases.

Additionally:
* Removes ActiveMailer, which we aren't using
* Turns "use_env" on which enables environment variable overrides of `config` gem configs (good in general, especially for heroku workflow)
* Adds [heroku workflow](https://psu.app.box.com/notes/578003316218)
* Upgraded bootsnap (not that important as it turned out, but I upgraded as part of a debugging process at one point)


cdmo:
I had to remove the bundler references in travis, otherwise it would just hang at a warning about a conflict, see [this failure](https://travis-ci.com/psu-libraries/myaccount/builds/141272054#L267)

I updated bundler to version 2 just because I thought it might quickly resolve things and because it needs to happen. But that didn't work.

Issue #20, opened 2019-11-13
Resolve login issues


cdmo:
The desired outcome:

- [x] when a user goes to any page of the app that user is sent to authentication (webaccess)
- [x] app should detect for session renewal need (stale?) and automatically renew session token
~[x] encrypt~

cdmo:
Re: encryption. This really doesn't make any sense as the auth info is passed as `GET` query params from the auth service. We would have to rewrite the auth service in order to hide these values.

cdmo:
Stale appears to be working correctly in my testing

ruthtillman:
Seems good.

PR #51, opened 2019-12-18
Swaps the order on warden authentication status check and current_use…

cdmo:
…r set/check

#20


app/views/summaries/index.html.erb
```diff
@@ -2,7 +2,6 @@
 
 <h3>
   Library account record for <%= patron.display_name %>
-  <%= link_to 'logout', logout_url, data: { turbolinks: false } if current_user? %>
```
banukutlu: thanks for removing this :)

PR #50, opened 2019-12-17
address rubygems 3.1.1 issue

cdmo:
#49 

## Additional change

Updated part of the README (just to trigger a build whilst debugging)

PR #54, opened 2019-12-19
Holds

cdmo:
For #5 adds viewing ability

Additional change:
* menu
* short time format update
* disabled a scss linter i didn't enjoy


cdmo:
Will rebase after #50 is merged

app/models/hold.rb
```diff
@@ -4,16 +4,62 @@
...
 
+  PICKUP_LOCATION = {
+    ABINGTON: 'Abington',
+    ACQ_DSL: 'Acquisitions (Law Schools)',
+    ACQUISTNS: 'Acquisitions (UP)',
+    ALTOONA: 'Altoona',
+    BEAVER: 'Beaver',
+    BEHREND: 'Erie',
+    BERKS: 'Berks',
+    BRANDYWINE: 'Brandywine',
```
banukutlu: @cdmo do we need double quotes here or not?

cdmo: Because there is a hyphen.. I guess I could have written this as a hash-rocket style. Or does an underscore get converted to a hyphen automatically? i can't remember...

banukutlu: oh I see 👍

cdmo:
Will do. Something's up with the build still too, will look into that. Thanks.

cdmo:
Added an issue to move logic out of that summaries partial https://github.com/psu-libraries/myaccount/issues/60

cdmo:
Will handle the conditional display of tables in the holds cancel branch

PR #59, opened 2020-01-06
renewals

banukutlu:
Renewals - part of #6 

### Not Included
- Select all functionality
- Custom error msgs #69

### DIFF
--- a/spec/controllers/renewals_controller.spec.rb
+++ b/spec/controllers/renewals_controller.spec.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe RenewalsController do
+  let(:user) do
+    { username: 'zzz123',
+      name: 'Zeke',
+      patron_key: '1234567',
+      session_token: 'e0b5e1a3e86a399112b9eb893daeacfd' }
+  end
+
+  let(:mock_patron) { instance_double(Patron, checkouts: checkouts) }
+  let(:checkouts) { [
+    instance_double(Checkout, item_key: '123', title: 'Renewal 1', non_renewal_reason: ''),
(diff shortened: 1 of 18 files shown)


spec/controllers/renewals_controller.spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  before do
+    warden.set_user(user)
+    allow(SymphonyClient).to receive(:new).and_return(mock_client)
+    allow(controller).to receive(:patron).and_return(mock_patron)
+  end
+
+  describe '#create' do
+    it 'requires list of checkouts to be renewed as params' do
+      expect { post :create, params: {} }.to raise_error(ActionController::ParameterMissing)
```
cdmo: Seems like we should add a scenario that triggers the `RenewalException`, right? I can do that manually by submitting a bad parameter (i.e., no match on the list of checkouts to the renew id).

By the way, I also wonder if `deny_access` should be `no_match_found`. Deny access feels wrong maybe?

banukutlu: oh yeah I think so, I will add a spec for `RenewalException`.

re `deny_access`: The way I thought of it was because the user does not have access to whatever item_key is being passed, it could be a valid item_key but since it is not checked out by that user, won't be able to renew. I am fine with either name, just sharing what I thought when I added that.

banukutlu: I totally forgot that I actually had a spec for it, here https://github.com/psu-libraries/myaccount/blob/a1bc58740958c2aafafe56bfce0ea0afc1e74bdc/spec/controllers/renewals_controller_spec.rb#L76. Sorry for the confusion :)

cdmo: well sure, but this deny_access will also be used when a bogus id is sent along. meh, not a big deal, :shipit: 

and yes, you're test at the end covers. 👍

Issue #64, opened 2020-01-10
Hide hold tables when there are no holds


cdmo:
Left over from #58

ruthtillman:
👍

PR #71, opened 2020-01-15
Adds conditional displays for holds

cdmo:
#64

### DIFF
--- a/app/views/holds/index.html.erb
+++ b/app/views/holds/index.html.erb
@@ -1,78 +1,16 @@
 <h1>Holds</h1>
 
-<%= form_with url: '/holds/multiple', method: :delete do %>
-  <table class="table">
-    <caption>Holds that are ready for pickup</caption>
-    <thead>
-    <tr>
-      <th scope="col">
-        Select <br>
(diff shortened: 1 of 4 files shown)


app/views/holds/index.html.erb
```diff
@@ -1,78 +1,16 @@
...
-      </tr>
-    <% end %>
-    </tbody>
+  <% if @holds_ready.present? %>
+    <%= render 'holds/holds_ready_table' %>
+  <% else %>
+    <p>You have no holds ready to be picked up at this time.</p>
+  <% end %>
+  <% if @holds_not_ready.present? %>
+    <%= render 'holds/holds_not_ready_table' %>
```
banukutlu: @cdmo no else needed for this case?

cdmo: No, because if you don't have any "hold not yet ready to be picked up" this doesn't seem to me to rise to the occasion of "we should let you know." If you have no holds of either type (ready, not ready) a message is displayed. I was just trying to play the scenarios out in my head. We can ask Ruth to weigh in, I was just kinda guessing.

banukutlu: OK, that makes sense to me, I think it is good like this. Ruth can review it on QA, would let us know if we need to let the user know :)

Issue #68, opened 2020-01-14
Hold cancel message


cdmo:
Make it mimic current hold cancel message (includes volume, and other things)

ruthtillman:
👍

PR #73, opened 2020-01-16
Updates hold cancelation success message to include call number.

cdmo:
#68


cdmo:
Would have loved to use ActionView Component here too, but, alas, use inside a Controller is not yet allowed, but should be soon. See https://github.com/github/actionview-component/issues/38

Issue #74, opened 2020-01-17
Renew All for checkouts


banukutlu:
Part of #6

ruthtillman:
Success! (I'm running into some issues w/accounts who have large numbers but I think this may be a heroku memory thing?)

cdmo:
@ruthtillman do you recall how many went through the request? I would like to take a quick double check.

ruthtillman:
It was one of the patrons with hundreds of records. I think they may have all renewed but then stalled.

PR #75, opened 2020-01-21
74 renew all

banukutlu:
Issue #74, also part of #6


spec/helpers/checkouts_helper_spec.rb
```diff
@@ -92,4 +93,13 @@
       expect(content).to include(renew_button)
     end
   end
+
+  describe '#render_renew_all' do
+    let(:content) { helper.render_renew_all }
+
+    it 'renders the right html' do
+      renew_all = '<input type="checkbox" name="renew_all" id="renew_all" value="1" data-select-all="renewal" />'
+      expect(content).to include(renew_all)
```
cdmo: Could get rid of line 98 and just have the expect look at `helper.render_renew_all`.  Kinda curious why you did it this way. 🤔

spec/helpers/checkouts_helper_spec.rb
```diff
@@ -78,7 +78,8 @@
 
     it 'renders the right html' do
       checkbox = '<input type="checkbox" name="renewal_list[]" id="renewal_list_" '
-      checkbox += "value=\"#{checkout.item_key}\" class=\"checkbox\" multiple=\"multiple\" />"
+      checkbox += "value=\"#{checkout.item_key}\" data-checkbox-type=\"renewal\" "
+      checkbox += 'class="checkbox" multiple="multiple" />'
```
cdmo: Do you like this better? I think the switching between single and double quotes is a little weird? Maybe? 🤷‍♂ 

```suggestion
            checkbox = "<input type=\"checkbox\" name=\"renewal_list[]\" id=\"renewal_list_\" "\
                 "value=\"#{checkout.item_key}\" data-checkbox-type=\"renewal\" class=\"checkbox\" "\
                 "multiple=\"multiple\" />"
```

spec/controllers/renewals_controller_spec.rb
```diff
@@ -27,6 +27,7 @@
     warden.set_user(user)
     allow(SymphonyClient).to receive(:new).and_return(mock_client)
     allow(controller).to receive(:patron).and_return(mock_patron)
+    stub_const('RenewalsController::RENEWAL_FLASH_LIMIT', 2)
```
cdmo: 😎

app/controllers/renewals_controller.rb
```diff
@@ -63,6 +70,12 @@ def bulk_renewal_flash(response, type)
       )
     end
 
+    def bulk_renewal_summary_flash(response, type)
+      return unless response[type].any?
+
+      flash[type] = I18n.t("renew_all_items_summary.#{type}_html", count: response[type].length)
+    end
+
```
cdmo: This is nice! I was just doing something similar in my holds branch. There might be a way to elevate this to ApplicationController and merge our two methods. But for now this is nice and 25 is a nice number to keep cookie overflows at bay. 👍

app/helpers/checkouts_helper.rb
```diff
@@ -20,10 +20,18 @@ def render_checkout_due_date(checkout)
   end
 
   def render_renewal_select(checkout)
-    check_box_tag 'renewal_list[]', checkout.item_key, false, class: 'checkbox', multiple: true
+    check_box_tag 'renewal_list[]', checkout.item_key, false,
```
cdmo: 🤔 wonder if it would make sense to combine functionality with `HoldsHelper::render_hold_select` (or at least part of it). Might be overkill just to avoid repeating a single line of code.

app/helpers/checkouts_helper.rb
```diff
@@ -20,10 +20,18 @@ def render_checkout_due_date(checkout)
...
+                  data: { checkbox_type: 'renewal' }, class: 'checkbox', multiple: true
   end
 
   def render_renew_button
     submit_tag 'Renew', class: 'btn btn-primary btn-renewable-submit'
   end
+
+  def render_renew_all
```
cdmo: So, for Holds, I didn't make a special helper for the select all special checkbox. Should I have?

spec/controllers/renewals_controller_spec.rb
banukutlu: 😎

app/helpers/checkouts_helper.rb
banukutlu: I was thinking about a similar thing when working on these, like select all as a component 🤔

spec/helpers/checkouts_helper_spec.rb
banukutlu: hmm I think I was just keeping it consistent with the other specs so not much a specific reason other than that, I can change it for sure, maybe will change the rest as well then 🤔 what do you think?

banukutlu: I guess it makes sense to use it like this when there are more than one example using it, and for this one, yes it does not much

banukutlu: so I will get rid of let for the ones that needs it once 👍

app/helpers/checkouts_helper.rb
banukutlu: hmm I do not think that you have to, but maybe we can chat about if we can refactor to reuse select all stuff in both checkouts and holds?

app/controllers/renewals_controller.rb
banukutlu: yes, I agree with merging or refactoring those two, let's chat about it 👍

spec/helpers/checkouts_helper_spec.rb
cdmo: I wasn't feeling too strong about this one way or another, but, if we did want to _always_ use `let` then I'd have to update some of my specs (for consistency). Thanks.

app/helpers/checkouts_helper.rb
cdmo: Oh man. Love that idea. I think for now lets leave as is and chat in Slack about making a new component.

app/helpers/checkouts_helper.rb
cdmo: Sounds good

Issue #62, opened 2020-01-10
Modify hold


cdmo:
Change hold pickup location and/or date no longer needed

Left over from #58

cdmo:
Just wanted to record the findings from Thursday: despite the ability to dynamically query all possible pickup locations for a patron and item combination, the list of pickup locations is set statically and is universal for all items and patrons.

cdmo:
@mkutch This is a note to remind you/us that we need the permissions changed on the production web services so that fillByDate can be edited by patrons.

cdmo:
@ruthtillman I have a UX question for you. In the current My Account the user encounters "confirm prompts" when attempting to cancel a hold - like "Are you sure you want to cancel these holds?" kind of thing. Do we want that in My Library?

ruthtillman:
I think as long as we don't have a "cancel all holds" button so they'd have to at least select individual or select All and confirm, it's ok (with caveat below). I would not say that for a public library, but we aren't the kind of place which stacks up holds on books. In public libraries, if there are 15 people in a queue and you're 4 and accidentally cancel and have to get back on the list, you won't see that book for _months_. Here, it's not as dire and we even have ILL.

BUT I think that does mean we need a solid response message with titles, etc. That way a user can go back to the catalog to place the hold again if they see they accidentally canceled one unintentionally. We should test with users.

cdmo:
Ok, cool. Well, here's a quick preview of a first attempt at creating a batch processing form section for holds:

![Screenshot 2020-01-21 10 33 12](https://user-images.githubusercontent.com/523381/72994581-59a41a80-3dc5-11ea-935c-72f1e4b83410.png)

One other nice thing is that we'll have a date picker widget "for free" with My Library because we're using the browser's native date picker (current My Account has a calendar icon that would seem to indicate a calendar widget, but it doesn't seem to work).

ruthtillman:
Yes, that's a very nice blue vs. red button and different sections to make it easier to edit than cancel. 👍

cdmo:
Hi @ruthtillman - @banukutlu brought up the following question during review of modifying holds:

> I was able to change the not needed after date without editing/selecting a pickup at, this seems to be not allowed in the current elibrary system. I just wanted to confirm that it is OK to update the not needed after date like this, which seems so to me 🤔

[...]

PR #76, opened 2020-01-21
Adds hold edit form details

cdmo:
#62 

* Also adds an extra spec for checking on item_details (private method)

Left the date changing stuff in there, and I plan on fleshing it out further in future PR. Or in this one if you'd rather, might depend on when we can expect to get an answer about using staff login.


app/models/hold.rb
```diff
@@ -104,6 +138,10 @@ def status_code
     fields['status']
   end
 
+  def bib_summary
+    "#{title} (#{call_number})"
```
banukutlu: good idea 👍 I should do the same for checkout

app/services/symphony_client.rb
```diff
@@ -28,6 +28,46 @@ def patron_info(user, item_details: {})
...
+
+  # not working!
+  # {
+  #   "messageList":[
+  #     {
+  #       "code": "permissionDeniedEditedFieldNotEditable",
+  #       "message": "PUT Permission denied: Field fillByDate not editable."
+  #     }
+  #   ]
+  # }
```
banukutlu: these can be deleted right now that it is working?

cdmo: oops sorry!

spec/models/hold_spec.rb
```diff
@@ -71,4 +71,11 @@
     hold.record['fields']['status'] = 'ACTIVE'
     expect(hold.status_code).to eq 'ACTIVE'
   end
+
+  it 'has a bib summary' do
+    hold.record['fields']['bib']['fields']['title'] = 'A wonderful title'
+    hold.record['fields']['item']['fields']['call']['fields']['dispCallNumber'] = 'ABC123'
+
+    expect(hold.bib_summary).to eq 'A wonderful title (ABC123)'
```
banukutlu: @cdmo wondering if there are items with no callnumber?

spec/controllers/holds_controller_spec.rb
```diff
@@ -34,8 +36,15 @@
       allow(mock_patron).to receive(:holds).and_return(holds)
     end
 
+    it 'sends the right item details to the web service' do
+      item_details = controller.send(:item_details)
+
+      expect(item_details).to eq holdRecordList: true
```
banukutlu: 👌

spec/models/hold_spec.rb
cdmo: we have items in our catalog without a call number? I can change if that's the case.

spec/controllers/holds_controller_spec.rb
cdmo: Don't love this actually, but I couldn't figure out a better way to test that private method. Just feel like that line is important to cover.

app/models/hold.rb
cdmo: might be better to send over to bib_record model

spec/controllers/holds_controller_spec.rb
banukutlu: per our conversation with Ruth, there should be always a call number in the response, so we are good 👍

cdmo: even if in the future there's a case where it doesnt, it would degrade pretty ok (just an empty parenthetical at the end of the title)

app/models/hold.rb
banukutlu: yes!

app/controllers/holds_controller.rb
```diff
@@ -47,7 +60,49 @@ def holds_not_ready
...
+      when 200
+        process_flash(:success, 'update_pickup.success_html')
+      else
+        Rails.logger.error(change_pickup_response.body)
+        process_flash(:error, 'update_pickup.error_html')
+      end
+    end
+
+    def handle_not_needed_after_request
+      raise HoldException, 'Error' if Date.parse(params['hold_expiration_date']) < Date.today
```
banukutlu: 👌

app/controllers/holds_controller.rb
```diff
@@ -47,7 +60,49 @@ def holds_not_ready
...
+      change_pickup_response = symphony_client.change_pickup_library(
+        hold_key: @hold_to_act_on.key,
+        pickup_library: params['pickup_library'],
+        session_token: current_user.session_token
+      )
+      case change_pickup_response.status
+      when 200
+        process_flash(:success, 'update_pickup.success_html')
+      else
+        Rails.logger.error(change_pickup_response.body)
```
banukutlu: nice, I should add one for renewals

spec/controllers/holds_controller_spec.rb
```diff
@@ -51,6 +60,67 @@
...
+        end
+
+        it 'and it\'s a date in the past, it does not update the not needed after and sets the flash message' do
+          patch :update, params: { id: 'multiple', hold_expiration_date: '2020-01-21', hold_list: [2] }
+
+          expect(flash[:error]).to match(/date that is in the past/)
+        end
+      end
+
+      context 'when not_needed_after param is sent and the webservice is responding with 200' do
```
banukutlu:
```suggestion
      context 'when not_needed_after param is sent and the webservice is not responding with 200' do
```
? 🤔

.rubocop_todo.yml
```diff
@@ -1,11 +1,23 @@
...
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 130
+
+# Offense count: 2
+# Cop supports --auto-correct.
+# Configuration parameters: AutoCorrect, AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
+# URISchemes: http, https
+Metrics/LineLength:
+  Max: 126
```
banukutlu: this todo seems to be not required anymore, @cdmo could you also check?

app/views/holds/_holds_not_ready_table.html.erb
```diff
@@ -35,5 +35,32 @@
...
+              <%= label_tag 'hold_expiration_date', 'Not needed after:' %>
+              <%= date_field_tag 'hold_expiration_date' %>
+              <p class="card-text"><small class="text-muted">The date you enter is the date that the system will stop trying to obtain the item for you.  It is not a guaranteed delivery date.</small></p>
+              <br>
+              <%= submit_tag('Update Selected Holds', class: 'btn btn-primary', formmethod: 'PATCH') %>
+            </div>
+          </div>
+          <div class="col-4 border-left">
+            <div class="card-body">
+              <h5 class="card-title">Cancel Selected Holds</h5>
```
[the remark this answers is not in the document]
cdmo: I think it adds a visual hierarchy that's needed, two halves of the panel have equal headings. [Ruth did review it in the issue and liked it](https://github.com/psu-libraries/myaccount/issues/62#issuecomment-577758899) fwiw. I understand where you're coming from from a pure logic standpoint, but I think from a design perspective it is better. Plus it's important  to really drive home that this is cancelling holds (a destructive action).

app/views/holds/_holds_not_ready_table.html.erb
```diff
@@ -35,5 +35,32 @@
...
+              <%= submit_tag('Cancel Selected Holds', class: 'btn btn-danger') %>
```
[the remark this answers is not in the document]
cdmo: Yes, I think so. Will change.

app/controllers/holds_controller.rb
```diff
@@ -11,22 +12,34 @@ def index
...
+  # PATCH /holds
+  # PUT /holds
+  def update
+    params['hold_list'].each do |holdkey|
+      @hold_to_act_on = holds.find { |hold| hold.key == holdkey }
+      handle_pickup_change_request if params['pickup_library'].present? && params['pickup_library'] != 'Not set'
+      handle_not_needed_after_request if params['hold_expiration_date'].present?
+    end
+
+    redirect_to holds_path
```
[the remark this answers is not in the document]
cdmo: Yes, this is true. 

[Asking Ruth](https://github.com/psu-libraries/myaccount/issues/62#issuecomment-578178789)

app/controllers/holds_controller.rb
cdmo: It's something I caught by mistake really, SUL didn't cover this problem.

app/controllers/holds_controller.rb
cdmo: :nods: 😄

.rubocop_todo.yml
cdmo: Nope.

app/views/holds/_holds_not_ready_table.html.erb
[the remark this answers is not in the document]
cdmo: Check latest

app/views/holds/_holds_not_ready_table.html.erb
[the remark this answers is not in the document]
banukutlu: I see your point, I guess I wished at least the text would be different or sth, but this is not my call and this is more of a next milestone thing rather than functionality so I'm fine with it and all good for this PR :)

app/views/holds/_holds_not_ready_table.html.erb
```diff
@@ -12,7 +12,7 @@
       <th scope="col">Item</th>
       <th scope="col">Availability</th>
       <th scope="col">Pickup at</th>
-      <th scope="col">Expires</th>
+      <th scope="col">Pick up by</th>
```
banukutlu: thanks, this is clearer :)

Issue #70, opened 2020-01-14
Do not link to PALCI records


ruthtillman:
PALCI records are not in the public catalog. When displaying, do not link.

cdmo:
@mkutch So, what is the canonical way to determine if an item is "of PALCI"? itemType or homeLocation?

ruthtillman:
Looks good.

cdmo:
I talked with @mkutch today, the itemType is canonical. She said that Shane Burris said that it is reliable 99.9% of the time, it is populated by a dynamic process. It is rare that this process fails and in those cases ILL will fix manually. (Just wanted to record this)

PR #79, opened 2020-01-24
Adds logic to not create a link if an item is from PALCI

cdmo:
#70 

actionview components made this pretty easy :)

### DIFF
--- a/app/components/bib_item_component.rb
+++ b/app/components/bib_item_component.rb
@@ -17,6 +17,10 @@ class BibItemComponent < ActionView::Component::Base
     "#{@title} / #{@author}"
   end
 
+  def final_title
+    link_to_unless @type == 'PALCI', processed_title, catalog_url
+  end
+
   def catalog_url
     CATALOG_URL + @catkey
   end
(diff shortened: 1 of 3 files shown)


app/components/bib_item_component.rb
```diff
@@ -17,6 +17,10 @@ def processed_title
     "#{@title} / #{@author}"
   end
 
+  def final_title
+    link_to_unless @type == 'PALCI', processed_title, catalog_url
```
banukutlu: nice 👌

spec/components/bib_item_component_spec.rb
```diff
@@ -8,14 +8,37 @@
...
     hold.record['fields']['bib']['fields']['title'] = 'A wonderful title'
     hold.record['fields']['bib']['fields']['author'] = ''
     result = render_inline(described_class, bibitem: hold)
-    expect(result.to_html).to include 'A wonderful title</a>'
+    expect(result.to_html).to include 'A wonderful title'
+  end
+
+  context 'when bib item type is anything but PALCI' do
+    before do
+      hold.record['fields']['item']['fields']['itemType']['key'] = 'ATLAS'
```
banukutlu: very cool format you used here @cdmo 😎

spec/components/bib_item_component_spec.rb
```diff
@@ -8,14 +8,37 @@
...
+    end
+
+    it 'renders an linked title' do
```
banukutlu:
```suggestion
    it 'renders a linked title' do
```
sorry for the totally unnecessary suggestion but I cannot help it :)

Issue #69, opened 2020-01-14
Custom error messages for renewals


banukutlu:
Map these error codes to the custom messages for renewals:

`messageList{code}=141`
Sirsi : `Override required to exceed renewal limit.`
Custom : `Denied -- Renewal limit reached, cannot be renewed.<br>Due date remains unchanged.`

`messageList{code}=7703`
Sirsi : `Override required to exceed unseen renewal limit.`
Custom :  `Denied -- Renewal limit reached, cannot be renewed.<br>Due date remains unchanged.`

`messageList{code}=105`
Sirsi : `Item has been recalled`
Custom : `Denied -- Item has been recalled, cannot be renewed.<br>Due date remains unchanged.`

`messageList{code}=252`
Sirsi : Item has holds
Custom : ` Denied -- Item has holds, cannot be renewed.<br>Due date remains unchanged.`

`messageList{code}=46`
Sirsi : `Item on reserve, may not be renewed`
Custom : `Denied -- Item on reserve, cannot be renewed.<br>Due date remains unchanged.`

For java errors, add the generic error msg and with a link to https://libraries.psu.edu/about/libraries

ruthtillman:
In revising overall error message:

- display shortened success/error messages if more than 25 items being renewed
- in both messages for error_html, change "Try again or contact us for help" to "Review list of checked out items below and contact us if you require assistance." link "contact us" to Ask a Librarian.

ruthtillman:
👍

PR #80, opened 2020-01-24
69 renewals error msgs

banukutlu:
Issue #69 

### Additional Changes

Refactoring to use bib_summary, added Rails.logger for failed renewals, handled an error when no items selected to renew

### DIFF
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -91,6 +91,6 @@ class RenewalsController < ApplicationController
     def error_prompt(error_message)
       return if error_message.empty?
 
-      content_tag(:p, "Denied: #{error_message}")
+      content_tag(:span, "<br>Denied: #{error_message}", nil, false)
     end
 end
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -62,7 +62,7 @@ class RenewalsController < ApplicationController
     def bulk_renewal_flash(response, type)
       return unless response[type].any?
 
-      flash[type] = I18n.t(
+      flash[type] = t(
         "renew_all_items.#{type}_html",
         count: response[type].length,
         items: content_tag(
@@ -80,7 +80,7 @@ class RenewalsController < ApplicationController
     def bulk_renewal_summary_flash(response, type)
       return unless response[type].any?
 
-      flash[type] = I18n.t("renew_all_items_summary.#{type}_html", count: response[type].length)
+      flash[type] = t("renew_all_items_summary.#{type}_html", count: response[type].length)
     end
 
     def renewal_prompt(renewal)
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -78,9 +78,10 @@ class RenewalsController < ApplicationController
 
     def renewal_prompt(renewal)
       if renewal.respond_to?(:each)
-        title_text = "#{renewal[0].title}, #{renewal[0].call_number}"
-        non_renewal_reason = error_prompt(renewal[1])
-        title_text + non_renewal_reason
+        renewal_obj, error_message = renewal
+        title_text = "#{renewal_obj.title}, #{renewal_obj.call_number}"
+        non_renewal_reason = error_prompt(error_message) unless error_message.empty?
+        title_text + (non_renewal_reason || '')
       else
         "#{renewal.title}, #{renewal.call_number}"
       end
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -87,11 +87,10 @@ class RenewalsController < ApplicationController
       if renewal.respond_to?(:each)
         renewal_obj, error_message = renewal
 
-        title_text = "#{renewal_obj.title}, #{renewal_obj.call_number}"
         non_renewal_reason = error_prompt(error_message) unless error_message.empty?
-        title_text + (non_renewal_reason || '')
+        renewal_obj.bib_summary + (non_renewal_reason || '')
       else
-        "#{renewal.title}, #{renewal.call_number}"
+        renewal.bib_summary
       end
     end
 
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -63,7 +63,7 @@ class RenewalsController < ApplicationController
       return unless response[type].any?
 
       items = response[type].map do |renewal|
-        content_tag(:li, renewal_prompt(renewal), {}, false)
+        content_tag(:li, renewal_attempt_report(renewal), {}, false)
       end
 
       items_list = content_tag :ul, safe_join(items, '')
@@ -77,15 +77,15 @@ class RenewalsController < ApplicationController
       flash[type] = t("renew_all_items_summary.#{type}_html", count: response[type].length)
     end
 
-    def renewal_prompt(renewal)
+    def renewal_attempt_report(renewal)
       if renewal.respond_to?(:each)
         renewal_obj, error_message = renewal
 
         non_renewal_reason = error_prompt error_message
-        renewal_obj.bib_summary + (non_renewal_reason || '')
-      else
-        renewal.bib_summary
+        return renewal_obj.bib_summary + (non_renewal_reason || '')
       end
+
+      renewal.bib_summary
     end
 
     def error_prompt(error_message)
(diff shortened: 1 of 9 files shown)


app/controllers/renewals_controller.rb
```diff
@@ -37,7 +38,13 @@ def checkouts_to_renew
...
 
     def renew_params
-      params.require(:renewal_list)
+      if params[:renewal_list].blank?
+        flash[:notice] = 'No items were selected for renewal.'
+
+        redirect_to checkouts_path
+      else
+        params[:renewal_list]
+      end
```
cdmo: I kinda feel like this should be a front-end validate, like unless a checkbox is checked don't allow the button to be clicked (i.e, disable it). What do you think? For the ruby code you can get rid of that `else`.

```suggestion
      end
      params[:renewal_list]

File: app/controllers/renewals_controller.rb [hunk 2]
```diff
@@ -37,7 +38,13 @@ def checkouts_to_renew
     end
 
     def renew_params
-      params.require(:renewal_list)
+      if params[:renewal_list].blank?
+        flash[:notice] = 'No items were selected for renewal.'
+
+        redirect_to checkouts_path
+      else
```
cdmo: I kinda feel like this should be a front-end validate, like unless a checkbox is checked don't allow the button to be clicked (i.e, disable it). What do you think? For the ruby code you can get rid of that else.

```suggestion
      end
      params[:renewal_list]

File: app/controllers/renewals_controller.rb [hunk 3]
```diff
@@ -78,15 +85,18 @@ def bulk_renewal_summary_flash(response, type)
...
-        title_text = "#{renewal[0].title}, #{renewal[0].call_number}"
-        non_renewal_reason = error_prompt(renewal[1])
-        title_text + non_renewal_reason
+        renewal_obj, error_message = renewal
+
+        non_renewal_reason = error_prompt error_message
+        renewal_obj.bib_summary + (non_renewal_reason || '')
       else
-        "#{renewal.title}, #{renewal.call_number}"
+        renewal.bib_summary
```
cdmo: 😎

app/models/concerns/bib_record.rb
```diff
@@ -81,6 +81,10 @@ def item_key
     fields.dig('item', 'key')
   end
 
+  def bib_summary
```
cdmo: ✔️

app/controllers/renewals_controller.rb
cdmo: Oh, and again here, you can turn the `else` into an `end` and remove the `end`

app/controllers/renewals_controller.rb
banukutlu: yeah, I want to add a front-end validate for this,that's why I brought up the form validations the other day, did not want to add front-end stuff yet before we talk about it. I do not know if it hurts to keep this check as well?

app/controllers/renewals_controller.rb
banukutlu: This is for to add the title to the message when there are no errors, the successful renewal case.

cdmo: Yup. So, the way I think of it is the top `if` is catching errors. The default scenario is when renewal.respond_to?(:each) ends up being false. and it's the only other scenario, so you can just remove the else...end and just leave it outside the above if that covers errors

app/controllers/renewals_controller.rb
cdmo: Oh sorry, I forgot that conversation. I can see pros and cons of doing it just front-end, just back-end or both. We can chat off GH of course. For now, just backend is in place and working so let's leave it. 👍

app/controllers/renewals_controller.rb
```diff
@@ -73,20 +80,23 @@ def bulk_renewal_flash(response, type)
...
       end
     end
 
     def error_prompt(error_message)
+      return if error_message.empty?
+
       content_tag(:div, "Denied: #{error_message}")
```
cdmo:
```suggestion
      content_tag(:p, "Denied: #{error_message}")
```
Sorry I didn't catch this earlier, but, can we wrap this in [a more semantic](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML) paragraph tag instead? Or does that break things?

spec/controllers/checkouts_controller_spec.rb
```diff
@@ -29,6 +29,12 @@
       allow(mock_patron).to receive(:checkouts).and_return(checkouts)
     end
 
+    it 'sends the right item details to the web service' do
```
cdmo: Thanks for adding this 👍

spec/controllers/renewals_controller_spec.rb
```diff
@@ -86,6 +94,16 @@
       end
     end
 
+    context 'when response include errored items with empty error message' do
+      let(:renew_items_response) { { success: [checkouts[0]], error: [[checkouts[2], '']] } }
+
+      it 'error messages include items title only if non renewal reason empty' do
```
cdmo:
```suggestion
      it 'error messages include items title only' do
```
Seems like `if non renewal reason empty` is redundant?

spec/services/symphony_client_spec.rb
```diff
@@ -100,11 +108,17 @@
                session_token: 'e0b5e1a3e86a399112b9eb893daeacfd')
     end
 
-    it 'returns successful + errored titles for individual renewal requests in symphony' do
-      renew_response = client.renew_items(user, checkouts)
+    it 'returns successful + errored items for individual renewal requests in symphony' do
```
cdmo:
```suggestion
    it 'returns all responses for individual renewal requests in symphony regardless of success or error' do
```

spec/controllers/renewals_controller_spec.rb
banukutlu: It happens when we get the java errors so I am not adding a detailed msg for the renewal flash, so there should be a div with text starting with `Denied:` in this case, that is what I was trying to test here but not sure not required to do so 🤔

app/controllers/renewals_controller.rb
banukutlu: If I understand correctly, you mean sth like this?

``` 
def renewal_prompt(renewal)
      if renewal.respond_to?(:each)
        renewal_obj, error_message = renewal

        non_renewal_reason = error_prompt error_message
        renewal_obj.bib_summary + (non_renewal_reason || '')
      end
    
      renewal.bib_summary
    end
```
For a successful renewal, `renewal` here only has the checkout obj but for failed ones it is an array of checkout and error msg, without the else in the failed renewal case `renewal.bib_summary` won't work 🤔  If I understand correctly or maybe I am confusing myself here :)

app/controllers/renewals_controller.rb
banukutlu: yes, let's chat about it and depending on what we decide how we will handle the validations in general, I can refactor this part again or remove etc.

spec/services/symphony_client_spec.rb
```diff
@@ -100,11 +108,17 @@
...
+      renew_response = client.renew_items(user, [checkouts.first, checkouts.second])
 
-      expect(renew_response).to include error: [[checkouts.last, 'Item has holds']],
+      expect(renew_response).to include error: [[checkouts.second, 'Item has holds']],
```
cdmo:
```suggestion
      expect(renew_response).to be error: [[checkouts.second, 'Item has holds']],
```

The two things being compared have equal (`eq`) values, is this more explicit/specifc?

app/controllers/renewals_controller.rb
banukutlu: yes, your suggestion is better 👍

spec/controllers/renewals_controller_spec.rb
cdmo: So, line 97 says `items with empty error message`

I thought that was the same thing?

app/controllers/renewals_controller.rb
banukutlu: refactored this per our conversation on slack

spec/controllers/renewals_controller_spec.rb
banukutlu: ahh I see, yes seems redundant

app/controllers/renewals_controller.rb
```diff
@@ -55,38 +62,33 @@ def deny_access
...
-          :ul,
-          safe_join(response[type].map do |renewal|
-            content_tag(:li,
-                        renewal_prompt(renewal),
-                        {},
-                        false)
-          end, '')
-        )
-      )
+      items = response[type].map do |renewal|
```
cdmo: It's even less lines of code! :) Thanks, to me this is so much clearer.

banukutlu: I agree! thanks!

Issue #60, opened 2020-01-06
Extract logic from _summaries.html.erb


cdmo:
Into helpers or whatever

ruthtillman:
Seems good!

PR #81, opened 2020-01-24
Refactors summaries, adds new ActionView Component for Dashboard Item

cdmo:
#60 

Also had to bump ActionView Components in order to get access to new `render?` method.

I did also try out rendering an additional ActionView Component from within an ActionView Component for the little label and number pairs, but it didn't seem worth it.

### DIFF
--- a/app/components/dashboard_item_component.rb
+++ b/app/components/dashboard_item_component.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class DashboardItemComponent < ActionView::Component::Base
+  def initialize(model:, count:, count_term: nil, items: nil)
+    @model = model
+    @count = count
+    @count_term = count_term
+    @items = items
+  end
+
+  def card_body
+    return nil if @items.nil?
+
+    facts = @items.map do |item|
+      content_tag :li, "#{item[:count]}  #{item[:label]}"
+    end.join
+
+    content_tag :ul, facts, nil, false
+  end
+
+  def item_link
+    url = url_for(controller: @model.to_s.downcase, action: 'index')
+
+    link_to "See all #{@model}", url, class: 'btn btn-primary float-right'
+  end
+
+  def render?
+    @count.positive?
+  end
+end
--- a/app/components/dashboard_item_component.rb
+++ b/app/components/dashboard_item_component.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 class DashboardItemComponent < ActionView::Component::Base
-  def initialize(model:, count:, count_term: nil, items: nil)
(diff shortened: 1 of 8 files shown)


cdmo:
Meant to add a preview 

Screenshot 2020-01-24 15 57 03

banukutlu:
> Meant to add a preview
> 
> Screenshot 2020-01-24 15 57 03

this looks so nice!

app/views/summaries/_summary.html.erb
```diff
@@ -6,36 +6,29 @@
 <% end %>
 
 <div class="page-section">
-  <h2>Summary</h2>
+  <h1>Summary</h1>
+  <div class="row">
+    <%= render DashboardItemComponent, model: 'Fines',
+                                       count: patron.fines.sum(&:owed_amount),
```
banukutlu:
```suggestion
                                       count: number_to_currency patron.fines.sum(&:owed_amount),
```

banukutlu: wondering if you removed `number_to_currency` on purpose?

app/components/dashboard_item_component.rb
```diff
@@ -0,0 +1,30 @@
...
+    content_tag :ul, facts, nil, false
+  end
+
+  def item_link
+    url = url_for(controller: @model.to_s.downcase, action: 'index')
+
+    link_to "See all #{@model}", url, class: 'btn btn-primary float-right'
+  end
+
+  def render?
```
banukutlu: 😎  very nice to be able to move this logic to the class

spec/components/dashboard_item_component_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+                                    count: 2
+                                  },
+                                  {
+                                    label: 'overdue',
+                                    count: 3
+                                  }
+                                ] }}
+
+  it 'makes a list' do
+    component = render_inline(described_class, dashboard_item_args).to_html
```
banukutlu: could we add component with a `let` like below and remove that each line from the examples?

`let(:component) { render_inline(described_class, dashboard_item_args).to_html }`

I tried it, seemed to be working but I might be missing sth as well? What do you think?

cdmo: yes, great idea. Also, I refactored reduce Mystery Guest

cdmo:
Now I remember why I didn't add `number_to_currency`! `count` needs to be an integer (or "positve-able". If one could easily coerce a currency into an integer (i.e., with `to_i`) then I wouldn't have needed to add a new argument: `currency`.

PR #84, opened 2020-01-27
Refactored to make the renewal response from Sirsi uniform.

cdmo:
I was thinking that we could make the renewal response uniform so that we can reduce the complexity of the code in the controller. 

A step further to me would be to turn this uniform hash into a small class like SirsiRenewalResponse.

### DIFF
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -79,7 +79,7 @@ class RenewalsController < ApplicationController
 
     def renewal_attempt_report(renewal_response)
       if renewal_response[:sirsi_response].present?
-        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} #{renewal_response[:sirsi_response]}"
+        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} Denied: #{renewal_response[:sirsi_response]}"
       end
 
       renewal_response[:renewal].bib_summary
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -79,7 +79,7 @@ class RenewalsController < ApplicationController
 
     def renewal_attempt_report(renewal_response)
       if renewal_response[:error_message].present?
-        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} #{renewal_response[:error_message]}"
+        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} Denied: #{renewal_response[:error_message]}"
       end
 
       renewal_response[:renewal].bib_summary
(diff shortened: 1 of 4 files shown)


cdmo:
Screenshot 2020-01-27 14 49 50

app/services/symphony_client.rb
```diff
@@ -76,7 +76,7 @@ def renew_items(user, checkouts)
 
       case response.status
       when 200
-        status[:success] << checkout
+        status[:success] << { renewal: checkout, error_message: nil }
```
banukutlu: what do you think for using just `message` instead of `error_message` for both, so a bit more generic?
not so important though, just a symbol

cdmo: I like it, much better. That was what I wrote in Slack initially 😉

app/controllers/renewals_controller.rb
```diff
@@ -78,17 +78,10 @@ def bulk_renewal_summary_flash(response, type)
     end
 
     def renewal_attempt_report(renewal_response)
-      if renewal_response.respond_to?(:each)
-        non_renewal_reason = error_prompt renewal_response[:error_message]
-        return renewal_response[:renewal].bib_summary + (non_renewal_reason || '')
+      if renewal_response[:error_message].present?
+        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} #{renewal_response[:error_message]}"
```
banukutlu:
```suggestion
        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} Denied: #{renewal_response[:error_message]}"
```

app/services/symphony_client.rb
cdmo: error_message is also taken in that it's a method, I was just kinda going off what you had put in the other branch. I'll update and push.

cdmo: updated in latest, used `sirsi_response` 🤷‍♂

app/controllers/renewals_controller.rb
```diff
@@ -78,17 +78,10 @@ def bulk_renewal_summary_flash(response, type)
     end
 
     def renewal_attempt_report(renewal_response)
-      if renewal_response.respond_to?(:each)
-        non_renewal_reason = error_prompt renewal_response[:error_message]
-        return renewal_response[:renewal].bib_summary + (non_renewal_reason || '')
+      if renewal_response[:sirsi_response].present?
+        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} #{renewal_response[:sirsi_response]}"
```
banukutlu:
```suggestion
        return "#{renewal_response[:renewal].bib_summary} #{tag(:br)} Denied: #{renewal_response[:sirsi_response]}"
```

cdmo: I thought I already committed this weird

app/services/symphony_client.rb
banukutlu: I liked sirsi_response better 👍

Issue #45, opened 2019-12-09
Add patron standing COLLECTION alert to summary page


banukutlu:
When a user is sent to collection this is the alert. The standing{key} is COLLECTION

![image](https://user-images.githubusercontent.com/43478453/70470879-063a6f00-1a9a-11ea-9d5f-ebbbbef82b42.png)

_Originally posted by @mkutch in https://github.com/psu-libraries/myaccount/issues/1#issuecomment-563427871_

ruthtillman:
Looks good.

PR #86, opened 2020-01-28
adding patron standing COLLECTION alert to summary

banukutlu:
Issue #45

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -53,8 +53,8 @@ class Patron
     record == { 'messageList' => [{ 'code' => 'sessionTimedOut', 'message' => 'The session has timed out.' }] }
   end
 
-  def standing_alert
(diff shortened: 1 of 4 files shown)


app/models/patron.rb
```diff
@@ -48,17 +53,17 @@ def stale?
...
+    PATRON_STANDING_ALERTS[standing] || ''
   end
 
   private
 
     def fields
       record['fields']
     end
+
+    def standing
```
cdmo: :nods:

cdmo: there is a pattern of doing `thing_code` and `thing_human` ... like item type in bib record and pickup library in hold and status in fines. Would that make sense here too?

banukutlu:
> Thoughts about using _code and _human?

yes, makes sense! I should remember to use those as standard :) I will update

Issue #83, opened 2020-01-24
Refactor select all as a component


banukutlu:

ruthtillman:
Can't judge refactors as much but this seems ok. The Select All box is in a weird place but I assume styling will take care of that.

PR #89, opened 2020-01-28
Select all component

banukutlu:
Issue #83

### DIFF
--- a/app/components/select_all_child_component.rb
+++ b/app/components/select_all_child_component.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class SelectAllChildComponent < ActionView::Component::Base
+  def initialize(type:, key:, name: nil)
+    @name = name
+    @key = key
+    @type = type
+  end
+
+  def final_name
+    @name ||= @type
+
+    "#{@name}_list[]"
+  end
+
+  private
+
+    attr_reader :name, :type, :key
+end
--- a/app/components/select_all_child_component.rb
+++ b/app/components/select_all_child_component.rb
@@ -1,19 +1,13 @@
 # frozen_string_literal: true
 
 class SelectAllChildComponent < ActionView::Component::Base
-  def initialize(type:, key:, name: nil)
-    @name = name
+  def initialize(name: nil, key:, target_keyword:)
+    @name = name || "#{target_keyword}_list[]"
     @key = key
-    @type = type
-  end
-
-  def final_name
-    @name ||= @type
-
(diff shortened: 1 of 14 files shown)


app/components/select_all_child_component.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def final_name
+    @name ||= @type
+
+    "#{@name}_list[]"
+  end
+
+  private
+
+    attr_reader :name, :type, :key
```
cdmo: So I guess the main reason to do this is to avoid calling instance variables directly. I like it. I should refactor my existing components because I don't think I included it.

banukutlu:
> I think this awesome, just one small suggestion to consider. It looks like with renewals you used `type` as `renewal` but with holds you used `hold` to be `name` and `pending` to be `type`, which was unexpected to me.
> 
> I think I would suggest either just using one argument called `keyword` (open to suggestions, maybe `target`?) to replace `name` and `type` (my preference) OR change `type` to `model` and use it consistently and add a `suffix`.
> 
> Thoughts?
> 
> And again, I'm loving these components.

Initially, I did not add a `name` argument and was just using type as a name with adding `_list[]` but then I realized holds actually using a `hold_list[]` as name for both ready and pending tables where the data attribute check is `pending` and `ready`.  Since I did not wanted to refactor holds so I ended up just adding an optional argument `name`. So, do you think it is OK to use different name for the child checkboxes for those two tables? Not sure how much change it would need for holds? 🤔

app/components/select_all_child_component.rb
banukutlu: Oh I actually got the idea from you here https://github.com/psu-libraries/myaccount/blob/475904401417f8783f47b6de746d1179153bf4a6/app/components/bib_item_component.rb#L30

cdmo: `:type` here refers to a bib record's type, like `BOOK` or `ATLAS` as opposed to `hold` or `checkout` which would be a model, right?

cdmo: I see ok, yeah, in dashboard item I didn't. Will make that change 👍

cdmo:
There's a refactor needed yes, it won't be bad. I'll do that.

banukutlu:
> There's a refactor needed yes, it won't be bad. I'll do that.

yeah, thanks, that would be great!

cdmo:
Ok, take a look at #93

cdmo:
:shipit:

Issue #85, opened 2020-01-27
Add conditional display to Bills


banukutlu:
Also change the table caption, or maybe remove?

PR #90, opened 2020-01-29
Hides fines table if there aren't any. (also fixed indentation issue …

cdmo:
…on holds index erb.

#85

### DIFF
--- a/app/views/fines/_fines_table.html.erb
+++ b/app/views/fines/_fines_table.html.erb
@@ -1,0 +1,28 @@
+<p>NOTE: Fines for items that are still checked out will not appear here until the items have been returned</p>
+<table class="table">
+  <caption>Bills Table</caption>
+  <thead>
+  <tr>
+    <th scope="col">Title</th>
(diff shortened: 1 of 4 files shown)


app/views/fines/_fines_table.html.erb
```diff
@@ -0,0 +1,28 @@
+<p>NOTE: Fines for items that are still checked out will not appear here until the items have been returned</p>
+<table class="table">
+  <caption>Bills Table</caption>
```
banukutlu:
```suggestion
  <caption>Bills</caption>
```

or maybe we even do not need the caption since there is also `<h1>Bills</h1>` ?

cdmo: I think this is outside the topic of what the issue is here. Maybe let's let #2 happen and make one sweep through after that is what I'm thinking. Thoughts?

banukutlu: sounds good 👍

PR #100, opened 2020-01-30
refactoring checkouts table into a partial

banukutlu:
also very minor update to hold and fine tables row, change from `<th>` to `<td>`


app/views/holds/_holds_ready_table.html.erb
```diff
@@ -15,12 +15,12 @@
     <tbody>
     <% @holds_ready.each do |hold| %>
       <tr>
-        <td>
+        <td scope="row">
           <%= render SelectAllChildComponent, key: hold.key, target_keyword: 'ready', name: 'hold_list[]' %>
         </td>
-        <th scope="row">
```
cdmo: Ok, this does make sense. Initially I'd coneived of the item summary being a kinda of table header (like a ["Table with two headers"](https://www.w3.org/WAI/tutorials/tables/two-headers/)) but it isn't 👍

app/views/holds/_holds_ready_table.html.erb
```diff
@@ -15,12 +15,12 @@
     <tbody>
     <% @holds_ready.each do |hold| %>
       <tr>
-        <td>
+        <td scope="row">
```
cdmo:
```suggestion
        <td>
```
[`scope` only applies to `th`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Advanced)

app/views/holds/_holds_not_ready_table.html.erb
```diff
@@ -19,12 +19,12 @@
     <tbody>
     <% @holds_not_ready.each do |hold| %>
       <tr>
-        <td>
+        <td scope="row">
```
cdmo:
```suggestion
        <td>
```

app/views/fines/_fines_table.html.erb
```diff
@@ -14,9 +14,9 @@
   <tbody>
   <% @fines.each do |fine| %>
     <tr>
-      <th scope="row">
+      <td scope="row">
```
cdmo:
```suggestion
      <td>
```

cdmo:
I'm waiting in suspense here @banukutlu 

> also very minor update to hold and fine tables row, change from to 

????? 😆 From what to what?

banukutlu:
> Looks good, will approve once my suggested changes are committed.
> 
> One idea too would be to move
> 
> `@checkouts.select(&:recalled?)`
> 
> into a method of `CheckoutsController` (like `recalled_checkouts`) and call it with a `.present?` in the erb file. Same with overdue. I did this with Holds.
> 
> Don't know if that really is much of an improvement though tbh.

sure, I would prefer the consistency, will do that way and commit

Issue #106, opened 2020-02-04
New hold form


cdmo:
from #96

cdmo:
@ruthtillman 

Do you have a preference in what we display to patrons in terms of items? Elibrary shows just the volume and I set My Account to show the whole call number. Now I'm not sure which we'd want. Here's a side by side:

![Screenshot 2020-02-11 09 53 40](https://user-images.githubusercontent.com/523381/74247774-9b353080-4cb4-11ea-8832-f958090ccd45.png)


Maybe we should just display the volume. The display call number for the current issue of the National Review is just "Periodical" which is confusing (not that a hold request for the current issue of the National Review would work anyway, but just saying)

ruthtillman:
Good question! Let's go with the volume segment. It takes up less screen real estate and the call number portion should almost never matter.

mkutch:
@ruthtillman can you take a look at this too? Their will be cases where there is no volume data.


![image](https://user-images.githubusercontent.com/43478453/74264775-3fc46c00-4ccf-11ea-9281-ac40eb66df5a.png)

ruthtillman:
Oh jeez, that is a good point @mkutch ... it also seems like the micro thing might be more useful than just no 29

I guess that's an argument to keep it. It would be too much overkill to do something v. detailed around it.

cdmo:
Re-opening, this still needs to be reviewed

ruthtillman:
Looking good. Re: the date picker, I think auto-populating (like we do in e-Library) it is probably a better idea than forcing the user to do so. Maybe 3 months out? When using it on my computer, one time I accidentally selected tomorrow and had to go back and fix it. Auto-populating and letting the user change seems like a friendlier choice for now.

cdmo:
@ruthtillman see #110. In e-library the default is 2 months from today so that's what we went with. I am deploying the code change (which hasn't been reviewed by @banukutlu just yet)  to heroku now so you can take a look.

PR #107, opened 2020-02-07
Adds HoldsController#new

cdmo:
#106 

Specs coming.


cdmo:
This latest commit handles #109 as well, see maryam's email

cdmo:
also #115

spec/services/symphony_client_spec.rb
```diff
@@ -124,4 +124,24 @@
       expect(renew_response).to include error: [error_response]
     end
   end
+
+  describe '#get_bib_info' do
+    before do
+      stub_request(:get, "#{Settings.symws.url}/catalog/bib/key/12345?includeFields=*,callList%7B*,itemList%7B*%7D%7D")
```
banukutlu: @cdmo just curious why you did not prefer to use `.with(query: hash_including(includeFields: '*,callList{*,itemList{*}}')`

app/models/bib.rb
```diff
@@ -0,0 +1,61 @@
...
+  # In the case where a holdable set is not discovered, nil is returned and the barcode instance attribute
+  # is available to process the hold request.
+  def holdables
+    return nil unless potential_holdables.count > 1
+
+    filter_holdables
+
+    return nil unless volumetric? && @holds.count > 1
+
+    @holds # .sort_by {|a, b| a.key <=> b.key }
```
banukutlu: @cdmo sort by will in another branch right?

cdmo: yes, remove this commented out sort, sorry

spec/services/symphony_client_spec.rb
cdmo: 🤷‍♂ no reason, not a thing I thought about. I'll change it

spec/models/bib_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  it 'will generate holdables when supplied with a body that has a callList' do
+    expect(bib.holdables.count).to be 8
+  end
+
+  it 'will show you an author' do
+    expect(bib.author).to be 'Hill Street blues (Television program)'
+  end
+
+  context 'with an item list of only one item' do
+    it 'will not generate any new holds' do
```
banukutlu: @cdmo no expect here?

cdmo: oops left 3 pending (without `xit`). Will add specs and commit.

spec/controllers/holds_controller_spec.rb
```diff
@@ -148,5 +148,23 @@
...
       end
     end
+
+    describe '#new' do
+      let(:bib) { build(:bib_with_holdables) }
+      let(:response_body) { HOLDABLE_LOCATIONS_RAW_JSON }
+
+      before do
+        stub_request(:get, 'https://example.com/symwsbc/catalog/bib/key/1?includeFields=*,callList%7B*,itemList%7B*%7D%7D')
+          .to_return(status: 200, body: bib.body.to_json, headers: {})
```
banukutlu: @cdmo `.with(query: hash_including(includeFields:''))` here as well? Works the same day but just a bit more readable? what do you think?

spec/controllers/holds_controller_spec.rb
```diff
@@ -148,5 +148,23 @@
...
+        stub_request(:get, 'https://example.com/symwsbc/policy/location/simpleQuery?includeFields=displayName,holdable&key=*')
```
banukutlu: @cdmo same as above ☝️

spec/controllers/holds_controller_spec.rb
cdmo: yes

spec/controllers/holds_controller_spec.rb
cdmo: ✔️

app/models/bib.rb
```diff
@@ -0,0 +1,61 @@
...
+        @holdable_locations.include? h.current_location
+      end
+    end
+
+    def volumetric?
+      @holds.find { |h| h.volumetric.present? }
+    end
+
+    def random_barcode
+      @body.dig('fields', 'callList')
```
banukutlu: @cdmo can this be `bib.dig('callList')`? does not matter much but just a bit shorter

cdmo: That's a good idea. I'll commit it. Mainly I added `def bib` so that it would be copacetic with bib_record (make `author` receive okay for example).

Aside: it does seem to me that `dig` is always preferable to specifying a key because key access risks an exception. I'll commit the change you brought up.

Issue #96, opened 2020-01-30
Ability to place holds


cdmo:
- [x] `Hold#create` - @banukutlu 
- [x]  Include symphony client request for placing holds - @banukutlu 
- [x] ~`Hold#new` form with pickup location and pickup up by date. Can use existing update hold not yet ready update form - need to turn into a partial and call from another view for new.~ see #106
- [x] Ability to handle redirect to where request came from (blackcat record show)
- [x] Explore concept of API end point for hold creation to be used by BlackCat

banukutlu:
@mkutch when placing holds with ws, it requires `SD-Working-LibraryID` header to be passed in. Would this be same as the pickup library or something else?

mkutch:
Talked with @banukutlu and the SD-Working-LibraryID is the patron's library.

cdmo:
@ruthtillman I am recommending we remove the "Reset" button here, all it does is reset the form (with two total inputs) back to their default state.

![Screenshot 2020-02-03 09 13 06](https://user-images.githubusercontent.com/523381/73660065-79f59400-4665-11ea-99ca-28ae73ec27b6.png)

I think it's more confusing than useful. Is that alright? We can keep "Cancel" and have it just take a user back to the catalog record they were in (as it does now)

ruthtillman:
I think cancel taking them back to the catalog record makes sense. Some forms are complex enough to require resetting, but this does not seem to be one of them.

cdmo:
this got autoclosed mistakenly, reopening

ruthtillman:
Tested and behaves well. One question -- if I go directly to something like http://myaccount-qa.herokuapp.com/holds/new?catkey=2890523 and hit Cancel, it doesn't seem to do anything.

cdmo:
Yep, that is note coded up yet. Will be functional early next week I think.

ruthtillman:
Getting 500 error during holds placement for JSON response but it's working.

banukutlu:
> Getting 500 error during holds placement for JSON response but it's working.

Fix for this issue added to https://github.com/psu-libraries/myaccount/pull/157

Issue #77, opened 2020-01-24
Cookie overflow avoidance app-wide, flash message protection


cdmo:
possible sol'n - [redis](https://medium.com/@kirill_shevch/configuration-cache-and-rails-session-store-with-redis-b3ce6f64d1fc)?

cdmo:
On hold pending resolving inconsistency in where requests get block processed

cdmo:
Wanted to make a note here that @banukutlu and I talked, and I believe we're going to avoid cookie overflow by not using cookies for messages and instead show highlighted rows for books that have been renewed (or whatever action)

cdmo:
marking this as closed as this is more of a technical thing that we can review rather than a front-facing change

PR #119, opened 2020-02-13
Adds HoldsController#create and HoldsController#result 

banukutlu:
part of Issue #96 and #133


app/models/hold.rb
```diff
@@ -5,7 +5,7 @@ class Hold
   include BibRecord
 
   PICKUP_LOCATION_REQUESTED = {
-    "UP-fPAT": 'Pattee Commons Services Desk',
```
cdmo: oops. 😆

config/locales/en.yml
```diff
@@ -62,4 +62,9 @@ en:
           past_date: You specified a "Not needed after" date that is in the past. "Not needed after" date must be in the future.
           pickup_choose_text: --Please choose a pickup location--
           success_html: <span class="font-weight-bold">Success!</span> "%{bib_summary}" pickup location was updated.<br>
-          error_html: <span class="font-weight-bold">Sorry!</span> Something went wrong and "%{bib_summary}" pickup location was not updated. Contact your campus library if you need assistance.<br>
\ No newline at end of file
+          error_html: <span class="font-weight-bold">Sorry!</span> Something went wrong and "%{bib_summary}" pickup location was not updated. Contact your campus library if you need assistance.<br>
+        place_hold:
+          missing_params: To place a hold "Pickup At library and "Not Needed After" date must be filled in.
+          select_volumes: Please select a volume to place a hold.
```
cdmo: We'll want to have this and other new stuff to be a part of the label review (i.e. #2)

cdmo: Might also be a good idea to review this in real time with Ruth to go through all of the scenarios that generate these messages.

app/controllers/holds_controller.rb
```diff
@@ -35,6 +37,60 @@ def new
     @bib = Bib.new(parsed_body, holdable_locations)
   end
 
+  # Handles placing holds
+  #
+  # POST /holds
+  def create
+    barcodes = params['barcode'].present? ? Array(params['barcode']) : params['barcodes']
+
+    raise HoldCreateException, 'Error' if barcodes.blank?
```
cdmo: Thinking one improvement would be to have a consistent plural `barcodes` for #create to expect to simplify this code and eliminate the extra check on whether or not a value is there (present? and blank?).

```rb
    raise HoldCreateException, 'Error' if params['barcodes'].blank?
    
    barcodes = [params['barcodes']].flatten
```

This would require changing `place_hold_hidden_input_component.html.erb` to

```
<%= hidden_field_tag :barcodes, bib.barcode %>
```

cdmo:
needs a rebase :)

cdmo:
So, one bad Idea I wanted to mention is to add a method to the Holds called `save` and have that validate that there's a barcode present and have it fire off a call to symphony client. I think this would more closely parallel ActiveRecord, but I don't think that's necessarily an expectation we want to set up anyway.

app/controllers/holds_controller.rb
```diff
@@ -122,4 +178,36 @@ def past_date
...
+                        t 'myaccount.hold.place_hold.missing_params'
+                      else
+                        t 'myaccount.hold.place_hold.select_volumes'
+                      end
+
+      redirect_to new_hold_path(catkey: params[:catkey])
+    end
+
+    def missing_args?
+      params['pickup_library'] === 'Not set' || params['pickup_by_date'].blank?
```
cdmo:
```suggestion
      params['pickup_library'] == 'Not set' || params['pickup_by_date'].blank?
```

So, using `===` kinda threw me here. The [String::=== method](https://ruby-doc.org/core-2.6/String.html#method-i-3D-3D-3D) just returns `==` with the added side affect of also passing true if the left side of this comparison is `String`.

Here's a weird thing that wouldn't happen but could:

```
irb(main):009:0> 'y'.class === 'x'
=> true
```

In JavaScript `===` is often preferred because `==` coerces to string, but ruby doesn't coerce at the `==` level of comparing. Also, looks like [eql? works the same as == in ruby](https://ruby-doc.org/core-2.6.3/Object.html#method-i-eql-3F).

```
irb(main):016:0> '1' == 1
=> false
irb(main):018:0> '1'.eql? 1
=> false
```

Sorry, too much on one little thing, just kinda threw me a bit.

app/controllers/holds_controller.rb
banukutlu: Yes, this would be great, I will add these changes in 👍

banukutlu:
> needs a rebase :)

yeah, I was thinking to rebase once I get your last PR in, if that is OK? So I do not need to rebase again.

app/controllers/holds_controller.rb
banukutlu: I do have a memory of why and how I used `===` here :) Yup, your point makes sense 👍

cdmo: What do you think about changing this method to a `case`? Right now it looks like "if there are missing args then tell the user they need to fill out location and pickup by day" else tell them to select a volume. But it seems like volume (i.e. barcode) is an argument right? I'm thinking (psuedo code) something like:

```
flash[:error] = case params
 when params['pickup_library'] === 'Not set'
   'error message'
 when params['pickup_by_date'].blank?
   'error message'
 when params['barcodes'].blank?
   'error message'
```

cdmo:
yup of course that's fine, good idea

app/controllers/holds_controller.rb
cdmo: And sorry again, not trying to be nitpicky or pedantic, I just kinda got down the rabbit hole on this one 🤣

app/controllers/holds_controller.rb
cdmo: This refactor is definitely about making hold work with new which were developed separately in parallel so there's bound to be stuff like this 👍

banukutlu: :100:

banukutlu: btw just committed the related changes 👍

banukutlu:
rebased now

app/controllers/holds_controller.rb
banukutlu: I will give this a try, I originally wanted to go with similar approach but can not recall right now why I ended up checking barcodes and other separately 🤔

cdmo: probably because of the whole `barcode` versus `barcodes` complexity?

banukutlu: made some changes for this and committed

app/controllers/holds_controller.rb
```diff
@@ -19,7 +21,7 @@ def index
   def update
     params['hold_list'].each do |holdkey|
       @hold_to_act_on = holds.find { |hold| hold.key == holdkey }
-      handle_pickup_change_request if params['pickup_library'].present? && params['pickup_library'] != 'Not set'
+      handle_pickup_change_request if params['pickup_library'].present? && params['pickup_library'].present?
```
cdmo: 😎

app/controllers/holds_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 class HoldsController < ApplicationController
   before_action :authenticate_user!
+  before_action :authorize_create!, only: :create
```
cdmo:
```suggestion
  before_action :check_for_blanks!, only: :create
```

I think this wording would be clearer about what's going on, I think of authorization to be individual-based checking of permissions. I was going to suggest "validate_params" but it's all blank checking so, seemed better?

banukutlu: changed 👍

app/controllers/holds_controller.rb
```diff
@@ -35,6 +37,56 @@ def new
...
+        status[:success] << { barcode: barcode,
+                              hold_key: JSON.parse(response.body).dig('holdRecord', 'key') }
+      else
+        Rails.logger.error("Place Hold for #{barcode} by #{patron.barcode}: #{response.body}")
+        status[:error] << { barcode: barcode,
+                            error_message: JSON.parse(response.body).dig('messageList')[0].dig('message') }
+      end
+    end
+
+    session[:place_hold_results] = results
```
cdmo: Cookie overflow still is an issue here, right? We're going to look at this across the board with #77 right? Starting to think adding a redis might be worth looking at. Take a look at [this blogpost](https://medium.com/@kirill_shevch/configuration-cache-and-rails-session-store-with-redis-b3ce6f64d1fc)?

app/controllers/holds_controller.rb
```diff
@@ -35,6 +37,56 @@ def new
...
+    session[:place_hold_catkey] = params['catkey']
+
+    redirect_to result_path
+  end
+
+  # Handles place hold response
+  #
+  # GET /holds/result
+  def result
+    (redirect_to holds_path) && return if session[:place_hold_results].blank?
```
cdmo:
```suggestion
    return redirect_to holds_path if session[:place_hold_results].blank?
```

I think this is a bit easier to read, the `&&` kinda through me tbh. Guess we could just `raise` here too, or, simply not validate at all and let a custom error page handle the error this will throw. It should be a pretty rare case that someone would go here directly. (like never?)

app/controllers/holds_controller.rb
banukutlu: yeah, would be an issue esp if we were to save the whole objects to the session, here I tried to go with the minimum info possible like hold key, barcode and error message if there is one. 

I will take a look at that blogpost, thanks for sharing.

app/controllers/holds_controller.rb
banukutlu: Happens when someone goes to /holds/result directly or if try to refresh the result page they are on, yes probably rare.

app/controllers/holds_controller.rb
cdmo: RIght, so it isn't likely to cause an overflow. But, in general, if we went with an alternative session store like Redis it would give us piece of mind across the application. Doesn't seem to difficult to implement either, just kinda drop-in with a gem and configure it.

banukutlu: Not likely but still we can not guarantee that it won't, I can think of a case for somebody trying to put a hold to many volumes of a record which might trigger the overflow issue. So yes, I agree that we need to look into Redis, also considering that this issue might happen for different actions across the application.

app/controllers/holds_controller.rb
```diff
@@ -35,6 +37,56 @@ def new
...
+    place_hold_results = session[:place_hold_results]
+    session.delete(:place_hold_results)
+
+    processed_results = place_hold_results.each do |status, results|
+      if status == 'success'
+        results.map { |result| result['placed_hold'] = hold_lookup(result['hold_key']) }
+      else
+        results.map { |result| result['failed_hold'] = item_lookup(result['barcode']) }
+      end
+    end
```
cdmo: This whole thing could be a candidate for extraction to a service object. I tried briefly, fwiw, to see if I could simplify, like 

```rb
    @place_hold_results = place_hold_results.map do |hold_result_group|
      hold_result_group.second.map {|hold_result| hold_lookup(hold_result['hold_key'])}
    end
```

but it was a dead end.

I do think you could just assign the instance variable on line 79 so it'd be 

```rb
     @place_hold_results = raw_place_hold_results.each do |status, results|
```

This would save a couple lines.

banukutlu: yes, will def. refactor controller to move out more logic out.

btw committed your suggestion, good idea 👍  thanks!

config/routes.rb
```diff
@@ -5,7 +5,9 @@
 
   resources :summaries, :fines, :checkouts, only: [:index]
   resources :renewals, only: [:create]
-  resources :holds, only: [:index, :new, :update, :destroy]
+  resources :holds, only: [:index, :new, :create, :update, :destroy]
+
+  match 'holds/result' => 'holds#result', via: :get, as: :result
```
cdmo:
```suggestion
  get 'holds/result', to: 'holds#result', as: :result
```
> In general, you should use the get, post, put, patch and delete methods to constrain a route to a particular verb
https://guides.rubyonrails.org/routing.html#http-verb-constraints

app/services/symphony_client.rb
```diff
@@ -93,6 +93,44 @@ def cancel_hold(holdkey, session_token)
                           })
   end
 
+  def place_hold(patron, session_token, item_barcode, hold_args)
+    body = {
+      "itemBarcode": item_barcode,
```
cdmo: surprised that rubocop didn't complain about double quotes here 🤔

app/services/symphony_client.rb
```diff
@@ -75,7 +75,7 @@ def renew_items(user, checkouts)
         status[:success] << { renewal: checkout, sirsi_response: nil }
       else
         Rails.logger.error("Renewal (#{checkout.item_key}): #{response.body}")
-        status[:error] << { renewal: checkout, sirsi_response: (error_message(response) || '') }
+        status[:error] << { renewal: checkout, sirsi_response: (renewal_error_message(response) || '') }
```
cdmo: I think this change is a nice improvement. I also think this could be moved to the controller to be consistent with the rest of our code, symphony client just being the thing that performs the request and the controller handles what to do with it. SymphonyClientResponse could even be a model maybe. But, this is not that important, just "thinking outloud" really.

banukutlu: yes agreed. I think such change for renewals would also be part of refactoring the holds_controller for another PR.

app/services/symphony_client.rb
banukutlu: oh yeah 🤔  idk why not, I will fix that.

cdmo: don't feel compelled, I'm just genuinely surprised

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
...
+      let(:place_hold_response) { instance_double(HTTP::Response) }
+      let(:mock_client) do
+        instance_double(SymphonyClient, place_hold: place_hold_response)
+      end
+
+      before do
+        allow(SymphonyClient).to receive(:new).and_return(mock_client)
+      end
+
+      context 'when placing hold to a not holdable item that has no volumes' do
```
cdmo:
```suggestion
      context 'when placing hold to an item that user already has a hold placed on' do
```

Also, I'm curious as to why you chose not to use the webmock approach here. Too much setup maybe?

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
...
+          post :create, params: place_hold_params
+
+          expect(session[:place_hold_catkey]).to eq '1'
+        end
+
+        it 'redirects to result page' do
+          post :create, params: place_hold_params
+
+          expect(response).to redirect_to result_path
+        end
```
cdmo: redundant?

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
...
+      end
+
+      context 'when placing hold to a holdable item with volumes' do
```
cdmo:
```suggestion
      context 'when placing multiple hold requests at once' do
```

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
...
+          post :create, params: place_hold_params
+
+          expect(session[:place_hold_catkey]).to eq '1'
+        end
+
+        it 'redirects to result page' do
+          post :create, params: place_hold_params
+
+          expect(response).to redirect_to result_path
+        end
```
cdmo: just not sure this has to be tested in separate scenarios, right? these two are the same code regardless of permutations of context. right?

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
...
+
+        it 'sets a flash error message' do
+          expect(flash[:error]).to match(/choose a not needed after date/)
+        end
+      end
+    end
+
+    describe '#result' do
+      let(:hold_info_response) { HOLD_LOOKUP_RAW_JSON.to_json }
```
cdmo:
```suggestion
      # HOLD_LOOKUP_RAW_JSON and ITEM_LOOKUP_RAW_JSON are mocked SymphonyClient responses
      let(:hold_info_response) { HOLD_LOOKUP_RAW_JSON.to_json }
```

spec/services/symphony_client_spec.rb
```diff
@@ -145,4 +145,135 @@
       expect(bib_response.body.to_str).to include('/catalog/bib')
     end
   end
+
+  describe '#place_hold' do
+    before do
+      stub_request(:post, uri)
```
cdmo: I see because you did the stub_requests here.

spec/controllers/holds_controller_spec.rb
```diff
@@ -169,5 +170,218 @@
         expect(assigns(:bib).holdables.count).to eq 8
       end
     end
+
+    describe '#create' do
+      let(:place_hold_params) {
+        { catkey: '1',
+          pickup_library: 'UP_PAT',
+          pickup_by_date: '02/02/2020' }
```
cdmo: should this be?

```suggestion
          pickup_by_date: '2020-02-02' }
```

spec/services/symphony_client_spec.rb
```diff
@@ -145,4 +145,135 @@
...
+    let(:hold_key) { 'a_hold_key' }
+    let(:uri) { "#{Settings.symws.url}/circulation/holdRecord/key/#{hold_key}" }
+    let(:include_fields) { '*,item{*,bib{title,author},call{*}}' }
+
+    let(:user) do
+      User.new(username: 'zzz123',
+               name: 'Zeke',
+               patron_key: 'some_patron_key',
+               session_token: 'e0b5e1a3e86a399112b9eb893daeacfd')
+    end
```
cdmo: Can you move one of these `let(:user)` statements up to the top of the file (under `let(:client)`)? And delete all the other let user statements? Will kill about 20-30 lines

spec/services/symphony_client_spec.rb
banukutlu: yes, that is why, and also my thinking was that the controller specs should not know about what the ws call is, so I did not use webmock approach in the controller but used it here for the client spec.

spec/factories/item.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+# frozen_string_literal => true
+
```
cdmo:
```suggestion
# frozen_string_literal: true
```

spec/services/symphony_client_spec.rb
banukutlu: sure, let me try, there might be one spec for unauthenticated user, I will add clear the user for that one.

spec/controllers/holds_controller_spec.rb
banukutlu: It might even be less setup actually but because I used stubs for the client spec and thought here controller just deals with what the client returns to it, not the request. Would like to hear your thoughts as well, I am open to ideas.

cdmo:
One other thing I was thinking is that the response components could be refactored into one, and perhaps encompass the surrounding markup too and can trigger off of 'error` or 'success` hashes being blank or not. But, not crucial.

spec/factories/item.rb
banukutlu: oops :)

banukutlu: why doesn't it letting me to commit the suggestion 🤔

spec/services/symphony_client_spec.rb
cdmo: you don't need to, `let` lazy loads...

spec/controllers/holds_controller_spec.rb
cdmo: Yup, makes sense 👍

spec/factories/item.rb
cdmo: 🤷‍♂ 
maybe because I already approved?

did like the new multi-line suggestion though 💯

banukutlu: ahhh yeah I guess so, I will commit it with the other changes 👍

cdmo: oddly my other suggestion seems commit-able 🤷‍♂

spec/services/symphony_client_spec.rb
banukutlu: commit coming in a bit :)

spec/controllers/holds_controller_spec.rb
banukutlu: yeah, this bugged me a bit either actually. I tried to not repeat those by nesting an upper level context but then rubocop nagged about the max # of nested context. Maybe you are right, there is no need to test them again in the other scenarios. so create action will redirect to result_path if there are some results(success or failed), and will redirect to holds_path if results are empty. Anyway I will just leave these checks in one context and remove from others

spec/factories/item.rb
banukutlu: yes 🤷‍♀  weird

spec/controllers/holds_controller_spec.rb
banukutlu: correct :) thanks!

banukutlu:
> One other thing I was thinking is that the response components could be refactored into one, and perhaps encompass the surrounding markup too and can trigger off of 'error` or 'success` hashes being blank or not. But, not crucial.

hmm yeah good idea, I will give it a try. First I started with that approach then I separated them i do not remember why :)

spec/factories/item.rb
banukutlu: ok committed the change

spec/controllers/holds_controller_spec.rb
cdmo: I think you could dump expectations on catkey and redirect in the main context... like before the contexts. Your call though, lmk what you want to do.

spec/controllers/holds_controller_spec.rb
banukutlu: yeah, removed it

Issue #108, opened 2020-02-10
Volume sort


cdmo:
Sort the checkboxes of items on the new hold form

PR #121, opened 2020-02-14
Adds a sort on holdables by naturalized volumetric (#108)

cdmo:
#108

### DIFF
--- a/app/models/bib.rb
+++ b/app/models/bib.rb
@@ -28,13 +28,31 @@ class Bib
 
     return nil unless volumetric? && @holds.count > 1
 
-    @holds
+    holds_sorted_naturally
   end
 
   private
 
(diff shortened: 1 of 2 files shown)


app/models/bib.rb
```diff
@@ -37,6 +37,24 @@ def bib
       @body['fields']
     end
 
+    def holds_sorted_naturally
+      @holds.each { |h| h.record['naturalized_volumetric'] = naturalize(h.volumetric.to_s) }
+        .sort_by { |h| h.record['naturalized_volumetric'] }
+    end
+
+    # Replace periods with spaces and then transform string into an array of "naturalized" values.
```
banukutlu: @cdmo thanks for the nice documentation here!

Issue #110, opened 2020-02-11
Default pickup by date 2 months from now


cdmo:
stemming from #106

PR #124, opened 2020-02-17
110 default pickup date

cdmo:
#110

#121 should be reviewed first.

### DIFF
--- a/app/views/holds/_form_inputs.html.erb
+++ b/app/views/holds/_form_inputs.html.erb
@@ -3,6 +3,6 @@
                                                     disabled: 'Not set', selected: 'Not set'),
                class: 'custom-select' %>
 <%= label_tag 'hold_expiration_date', 'Not needed after:' %>
-<%= date_field(:hold, :pickup_by_date) %>
+<%= date_field :hold, :pickup_by_date, value: default_pickup_by_date %>
 <p class="card-text"><small class="text-muted">The date you enter is the date that the system will stop
   trying to obtain the item for you.  It is not a guaranteed delivery date.</small></p>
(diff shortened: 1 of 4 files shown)


app/views/holds/_form_inputs.html.erb
```diff
@@ -3,6 +3,6 @@
                                                     disabled: 'Not set', selected: 'Not set'),
                class: 'custom-select' %>
 <%= label_tag 'hold_expiration_date', 'Not needed after:' %>
-<%= date_field(:hold, :pickup_by_date) %>
+<%= date_field :hold, :pickup_by_date, value: default_pickup_by_date %>
```
banukutlu: @cdmo Would you switch to `date_field_tag` instead of `date_field`?  The former was easier to use when creating holds. I have that changed in #119 PR but this is gonna merged before that one ;)

spec/helpers/holds_helper_spec.rb
```diff
@@ -96,4 +98,21 @@
       end
     end
   end
+
+  describe '#default_pickup_by_date' do
+    before do
+      # Instantiate two instances of DateTime before it is doubled.
+      arbitrary_datetime
+      arbitrary_datetime_plus
+      date_time = class_double('DateTime')
```
banukutlu: 👀 🆒

app/views/holds/_form_inputs.html.erb
cdmo: So... I did change it at https://github.com/psu-libraries/myaccount/pull/128/files#diff-af88dbf8fe522d62f05ab051e404c44eR6

Sorry, maybe overlook for now though?

spec/helpers/holds_helper_spec.rb
cdmo: if you don't instantiate the two arbitrary date times then the class double prevents them from being created, so order is important

app/views/holds/_form_inputs.html.erb
banukutlu: ahhh sure 👍

cdmo:
danke. #125 has been rebased too

PR #125, opened 2020-02-17
Generic actionview component

cdmo:
This is my own idea, based on a small bug where we saw "Author: " and no value.

Screenshot 2020-02-17 15 27 33

#121 and #124 should be reviewed first.

### DIFF
--- a/app/components/generic_value_with_label_component.rb
+++ b/app/components/generic_value_with_label_component.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class GenericValueWithLabelComponent < ActionView::Component::Base
+  def initialize(label:, value:)
+    @label = label
+    @value = value
+  end
+
+  def render?
+    @value.present?
+  end
+
+  private
+
+    attr_reader :label, :value
+end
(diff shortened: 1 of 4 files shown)


app/components/generic_value_with_label_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class GenericValueWithLabelComponent < ActionView::Component::Base
```
banukutlu: @cdmo I really like this idea 👍

spec/components/generic_value_with_label_component_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe GenericValueWithLabelComponent, type: :component do
+  let(:value) { 'Eleventy Billion' }
+  let(:label) { 'Number of Insects' }
```
banukutlu: 😄

cdmo:
ALL TEH COMPONENTZ!!11!!1

(#128 is rebased 🎂)

Issue #126, opened 2020-02-17
Front end validation for the form hold#new


cdmo:

ruthtillman:
Date picker validates. Even if I try to manually input a new date that's before today, it fails.

PR #128, opened 2020-02-18
126 frontend validation

cdmo:
#126

#121, #124, #125 should be reviewed first.


spec/views/holds/_form_inputs.html.erb_spec.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'holds/_form_inputs.html.erb', type: :view do
+  let(:bib) { build(:bib_with_holdables) }
+  let(:page) { Capybara.string(rendered) }
```
banukutlu: 👍

PR #129, opened 2020-02-18
Adds Phusion Passenger to be used specifically in local development t…

cdmo:
…o more closely mirror production.

This is a drop in replacement for local puma. See this [Phusion Quickstart guide](https://www.phusionpassenger.com/docs/tutorials/quickstart/ruby/)

Not sure this is worth it honestly.


cdmo:
I still think this is a good idea for more relative consistency with production.

PR #135, opened 2020-02-24
Service class for Form Builder and item type mapping on volumetrics

cdmo:
…formation building.

I don't want a review, just a glance over


cdmo:
This WIP does have itemType human added too

Screenshot 2020-02-24 15 54 50

banukutlu:
> …formation building.
> 
> I don't want a review, just a glance over

We actually chat about this PR but I wanted to make a note here as well. I liked your approach for the refactoring here, esp for moving logic out off the controller to the service object. Other than couple minor naming things that we talked about, I think this approach is 💯

banukutlu:
> …formation building.
> 
> I don't want a review, just a glance over

We already chat about this PR but I want to make a note here as well. I like that you refactored to move logic out off the controller to the service object and added a call model. Other than couple minor naming things that we talked about, I think this approach is 💯

cdmo:
K, this is 90% done, missing a few specs for builder, but all else should be there. Did not run rubo on it yet so I'm sure I've many violations so Travis will fail.

cdmo:
Just finished this up quick. Couple notes:

1. Thinking we should tell rubocop to ignore support files
2. This does not have grouping done, just the item type map and refactor to a service class.

app/components/place_hold_checkbox_wrapper_component.rb
```diff
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 
 class PlaceHoldCheckboxWrapperComponent < ActionView::Component::Base
-  def initialize(holdables: nil)
-    @holdables = holdables
+  def initialize(volumetric_calls: nil)
+    @volumetric_calls = volumetric_calls
```
banukutlu: 👌 liked volumetric_class better than holdables, nice!

app/services/place_hold_form/builder.rb
```diff
@@ -0,0 +1,87 @@
...
+      title: bib_info.title,
+      author: bib_info.author,
+      volumetric_calls: @volumetric_calls,
+      barcode: @volumetric_calls.present? ? nil : @call_list.sample.items.sample.barcode
+    }
+  end
+
+  private
+
+    def parsed_symphony_response(symphony_call, *params)
```
banukutlu: @cdmo What would you think about moving this to `symphony_client`? I think there are potential usage for it other the form builder and it is related to symphony so moving to the symphony service seems to make sense to me but want to hear your thoughts?

app/services/place_hold_form/builder.rb
```diff
@@ -0,0 +1,87 @@
...
+    @catkey = catkey
+    @user_token = user_token
+    @client = client # SymphonyClient
+    @volumetric_calls = []
+  end
+
+  def generate
+    bib_info = Bib.new(parsed_symphony_response(:get_bib_info, @catkey, @user_token))
+    @call_list = bib_info.call_list.map { |c| Call.new record: c }
+    @holdable_locations = find_holdable_locations
```
banukutlu:
```suggestion
    find_holdable_locations
```

What about setting `@holdable_locations` with find_holdable_locations? Similar to `process_volumetric_calls`?

banukutlu: It seems like `@holdable_locations` is used only for `filter_holdables` (if I am not missing sth), maybe then getting rid of line 15 here and just doing below would work?

```
 def filter_holdables
      holdable_locations = find_holdable_locations
      @call_list.filter! do |c|
        current_locations = c.items.map(&:current_location)
        current_locations.filter { |loc| holdable_locations.include? loc }.present?
      end
    end
```

app/services/place_hold_form/builder.rb
```diff
@@ -0,0 +1,87 @@
...
+      all_locations.filter { |p| p&.dig 'fields', 'holdable' }
+        .map { |p| p&.dig 'key' }
+    end
+
+    def filter_holdables
+      @call_list.filter! do |c|
+        current_locations = c.items.map(&:current_location)
+        current_locations.filter { |loc| @holdable_locations.include? loc }.present?
+      end
+    end
```
banukutlu:
```suggestion
    def filter_holdables
      holdable_locations = find_holdable_locations
      @call_list.filter! do |c|
        current_locations = c.items.map(&:current_location)
        current_locations.filter { |loc| holdable_locations.include? loc }.present?
      end
    end
```

spec/components/place_hold_hidden_input_component_spec.rb
```diff
@@ -3,16 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe PlaceHoldHiddenInputComponent, type: :component do
-  let(:bib_with_holdables) { build(:bib_with_holdables) }
-  let(:bib_without_holdables) { build(:bib_without_holdables) }
-  let(:component) { render_inline(described_class, bib: bib_without_holdables).to_html }
-
-  it 'renders when there are one or no holdable items to request' do
+  it 'renders when there\s a barcode' do
```
banukutlu:
```suggestion
  it 'renders when there\'s a barcode' do
```

spec/controllers/holds_controller_spec.rb
```diff
@@ -151,23 +151,23 @@
...
       before do
-        stub_request(:get, 'https://example.com/symwsbc/catalog/bib/key/1')
-          .with(query: hash_including(includeFields: match(/\*,callList/)))
-          .to_return(status: 200, body: bib.body.to_json, headers: {})
-
-        stub_request(:get, 'https://example.com/symwsbc/policy/location/simpleQuery')
-          .with(query: hash_including(includeFields: match(/displayName,holdable/)))
-          .to_return(status: 200, body: response_body.to_json, headers: {})
+        allow(PlaceHoldForm::Builder).to receive(:new).and_return(form_builder)
+        allow(form_builder).to receive(:generate).and_return(form_params)
```
banukutlu: 👌

spec/views/holds/new.html.erb_spec.rb
```diff
@@ -3,17 +3,43 @@
...
 
-  context 'with a new hold with holdables to choose' do
+  context 'without volumetric holdables to choose' do
     before do
-      assign(:bib, bib)
+      assign(:place_hold_form_params, form_params)
     end
 
-    it 'renders an input with a unique id attribute by barcode' do
+    it 'does not render checkbox inputs when there' do
```
banukutlu:
```suggestion
    it 'does not render checkbox inputs' do
```

spec/services/place_hold_form/builder_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    context 'when there aren\'t enough volumetric calls to present to require user select' do
+      let(:bib_info) { build(:bib_without_holdables) }
+
+      it 'will not generate any volumetric holdables' do
+        expect(form_params[:volumetric_calls]).to be_empty
+      end
+
+      it 'will pass along a random barcode' do
+        # @call_list.collect {|c| c.items.collect { |i| i.barcode }}.flatten
```
banukutlu:
```suggestion
```

app/components/place_hold_checkbox_wrapper_component.rb
cdmo: It's definitely more accurate and uses the terminology of the domain rather than inventing a new term.

app/services/place_hold_form/builder.rb
cdmo: I like this idea very much. I think I'd almost rather it live in it's own stand-alone class like `SymphonyClient::Parse` or `SymphonyClientParser` or something like that? Would like to handle in a separate issue though...

app/services/place_hold_form/builder.rb
cdmo: I think that's a cleaner code version, but, it would likely be less efficient because that `holdable_locations` method would get called many times rather than just once, right?

spec/controllers/holds_controller_spec.rb
cdmo: 😎

spec/views/holds/new.html.erb_spec.rb
cdmo: thanks, I should've read over the plain english specs again I guess! 🙃

spec/services/place_hold_form/builder_spec.rb
cdmo: Believe it or not I intended on leaving this here so I could illustrate where that array of barcodes came from. I'll add a line to explain.

cdmo: Also, randomness testing is imprecise :)

app/services/place_hold_form/builder.rb
banukutlu: `filter_holdables` called by `process_volumetric_calls` which is called by `generate`, so I thought it would not be many times but maybe I missing sth?

app/services/place_hold_form/builder.rb
banukutlu: yeah sure both works, I was thinking I can use that to refactor [hold_lookup](https://github.com/psu-libraries/myaccount/blob/94c2549dfe3bf4ea4dfb8085bfb55aebb1735427/app/controllers/holds_controller.rb#L176) and [item_lookup](https://github.com/psu-libraries/myaccount/blob/94c2549dfe3bf4ea4dfb8085bfb55aebb1735427/app/controllers/holds_controller.rb#L182) :)

cdmo: Cool. We can extract this method when we refactor one of those then.  There are a few other places too we could use this, like `Hold#hold_lookup`.  Not urgent by any means I think.

app/services/place_hold_form/builder.rb
cdmo: Okay, yep I see. I do like this better I think. I'll make the change. If there's one thing this gives up it's that the need to conduct this extra external class call (i.e., dependency) is a bit buried more. But, I think it's okay and this is more efficient.

.rubocop_todo.yml
```diff
@@ -11,6 +11,18 @@
 Metrics/ClassLength:
   Max: 184
 
-# Offense count: 8
+# Offense count: 1
+# Cop supports --auto-correct.
+# Configuration parameters: AutoCorrect, AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
+# URISchemes: http, https
+Metrics/LineLength:
+  Max: 91424
```
banukutlu: yeah shall we ignore this? move it to config rather than todo?

cdmo: Yeah, I think it makes sense to ignore it altogether. That ok? Might have to move the factory_bot config because we don't want that ignored.

banukutlu: yeah, makes sense 👍

cdmo: :nods: added

banukutlu: thanks!

Issue #137, opened 2020-02-26
Custom error response pages


banukutlu:

cdmo:
like page not found and 500s?

banukutlu:
> like page not found and 500s?

yup

ruthtillman:
Note -- when placing hold and it had JSON/500 error, no error message.

PR #139, opened 2020-02-27
custom error msgs for 404 and 500s

banukutlu:
Issue #137 

The text for error messages could be revised as part of #2


spec/requests/errors_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  describe 'internal server error' do
+    before (:all) { get '/500' }
+
+    it 'has http status 500' do
+      expect(response).to have_http_status(:internal_server_error)
+    end
+
+    it 'redirects to customized internal_server_error error page' do
+      expect(response.body).to include("We're sorry, but something went wrong.")
+    end
```
cdmo: This would pass even if the error wasn't customized, right? Would there be a way to see that this is being processed by the custom partial (maybe checking the partial used and matches?).

Also this also only covers `get`, wonder if we should add a `post` at least to test that?

config/routes.rb
```diff
@@ -10,4 +10,12 @@
...
 
   get '/logout', to: 'sessions#destroy', as: :logout
+
+  # error pages
+  match '/404' => 'errors#not_found', via: :all
+  match '/422' => 'errors#not_found', via: :all
+  match '/500' => 'errors#internal_server_error', via: :all
+
+  # catchall for not predefined requests - keep this at the very bottom of the routes file
+  match '*catch_unknown_routes' => 'errors#not_found', via: :all
```
cdmo: Avoids hash rocket (surprised that rubo didn't pick this up) and uses more recent `to:` syntax (guessing it's more recent)

```suggestion
  # error pages
  match '404', to: 'errors#not_found', via: :all
  match '422', to: 'errors#not_found', via: :all
  match '500', to: 'errors#internal_server_error', via: :all

  # catchall for not predefined requests - keep this at the very bottom of the routes file
  match '*catch_unknown_routes', to: 'errors#not_found', via: :all
```

banukutlu:
> This looks good to me now thanks.
> 
> I'm just curious, if you had left the default error message templates in public.. would the custom still show up? Seems like it would.
> 

Yes, I tried and afair nope custom did not show up in that case with public ones in, it first looks at the public folder I think.

cdmo:
Yeah, no, the `422` is in the defaults too, just seems like a really odd http response to make sure to build a response to in Rails proper. 🤷‍♂ 

Thanks!

Issue #120, opened 2020-02-14
Workaround for records in use response


banukutlu:
Symphony responses with `"The records are currently in use.  Please try again later."` error msg when canceling or placing multiple holds, and seems like after these actions since the record is busy, item info can not be retrieved as well.

We might have to take a try again approach as a workaround.

cdmo:
I believe @mkutch put in a ticket with Sirsi on this. Any response?

mkutch:
Sirsi hasn't been able to replicate the issue in-house yet. They asked for some info which I responded.

cdmo:
Ok. I think then we'll wait until that thread resolves before jumping in to fix this on our end then.

cdmo:
@mkutch and I worked to create reproducible steps for the vendor, the vendor verified the issue and acknowledged it's happening for them too. Their response is that we must just send repeat queries (as we have suspected). So this issue can be worked on now.

ruthtillman:
I don't know exactly how to test this, but reading through the process looks good.

PR #147, opened 2020-02-28
fix for records in use issue

banukutlu:
Issue #120


cdmo:
You see this didn't pass CI?

app/services/symphony_client_parser.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Responsible for parsing SymphonyClient responses.
+class SymphonyClientParser
+  def self.parsed_response(symphony_client, symphony_call, *params)
+    client_response = symphony_client.send(symphony_call, *params)
+    JSON.parse client_response.body
```
cdmo: Yup, this is great. 🌞

banukutlu: 😎

app/controllers/holds_controller.rb
```diff
@@ -167,14 +167,21 @@ def past_date
     end
```
cdmo: I think this could use a comment because it's kinda weird.

```suggestion
    # Fetch a hold lookup as many times as necessary up to 5 seconds to avoid `nil` item in response. 
```

app/controllers/holds_controller.rb
```diff
@@ -167,14 +167,21 @@ def past_date
...
 
     def hold_lookup(hold_key)
-      hold_record = symphony_client.get_hold_info hold_key, current_user.session_token
-      parsed_hold = JSON.parse hold_record.body
+      parsed_hold = SymphonyClientParser::parsed_response(symphony_client,
+                                                          :get_hold_info, hold_key, current_user.session_token)
+
+      start = DateTime.now
+      while parsed_hold.dig('fields', 'item').nil? && start + 5.seconds > DateTime.now
```
cdmo: Hmm, the new global delay constant doesn't seem to get called here. I think I expected that it would be. Is there a reason it isn't?

spec/services/symphony_client_spec.rb
```diff
@@ -127,7 +127,7 @@
     it 'returns the Symphony Client "catalog bib" resource type' do
       bib_response = client.get_bib_info '12345', user.session_token
 
-      expect(bib_response.body.to_str).to include('/catalog/bib')
+      expect(bib_response.body.to_str).to include '/catalog/bib'
```
cdmo: Who needs parens! 🔥

banukutlu: :metal:

app/services/symphony_client.rb
```diff
@@ -163,15 +162,20 @@ def renew_item_request(resource, item_key, headers: {})
                             })
     end
 
-    def error_message(response)
-      JSON.parse(response.body).dig('messageList')[0].dig('message')
+    def error_code(response)
+      return if response.status.ok?
+
+      JSON.parse(response.body).dig('messageList')[0].dig('code')
```
cdmo: Probably could make a little mini model out of `SymphonyResponse`. Just thinking outloud. Not a request for change at all.

spec/controllers/holds_controller_spec.rb
```diff
@@ -350,14 +354,38 @@
         get :result, params: {}, session: { place_hold_catkey: '1', place_hold_results: results }
 
         placed_hold = assigns(:place_hold_results)[:success].first[:placed_hold]
-        expect(placed_hold.record.first['key']).to eq 'a_hold_key'
+        expect(placed_hold.record['key']).to eq 'a_hold_key'
```
cdmo: Nice fix.

spec/controllers/holds_controller_spec.rb
```diff
@@ -350,14 +354,38 @@
...
       it 'assigns failed holds results correctly' do
         get :result, params: {}, session: { place_hold_catkey: '1', place_hold_results: results }
 
         failed_hold = assigns(:place_hold_results)[:error].first[:failed_hold]
-        expect(failed_hold.record.first['fields']['barcode']).to eq 'not_holdable_barcode'
+        expect(failed_hold.record['fields']['barcode']).to eq 'not_holdable_barcode'
+      end
+
+      context 'when hold lookup does not return item level info' do
+        let(:hold_info_empty_item_response) { HOLD_LOOKUP_NIL_ITEM_RAW_JSON.to_json }
```
cdmo: Think this is worth it? Follows pattern from 318

```suggestion
      context 'when hold lookup does not return item level info' do
        # HOLD_LOOKUP_NIL_ITEM_RAW_JSON is a mocked SymphonyClient response
        let(:hold_info_empty_item_response) { HOLD_LOOKUP_NIL_ITEM_RAW_JSON.to_json }
```

spec/controllers/holds_controller_spec.rb
```diff
@@ -350,14 +354,38 @@
...
+        let(:hold_info_params) { ['a_hold_key', user[:session_token]] }
+        let(:mock_client) { instance_double(SymphonyClient) }
+
+        let(:results) { {
+          success: [{ barcode: 'holdable_barcode', hold_key: 'a_hold_key' }]
+        }.with_indifferent_access }
+
+        before do
+          allow(SymphonyClientParser).to receive(:parsed_response).with(mock_client, :get_hold_info, *hold_info_params)
+            .and_return(parsed_hold_info_empty_item, parsed_hold_info)
```
cdmo: So this is saying that the `SymphonyClientParser` would return two objects when `get_hold_info` is invoked? Am I reading that right? Is that possible? Wouldn't it always return just one hold record?

spec/controllers/holds_controller_spec.rb
```diff
@@ -350,14 +354,38 @@
...
+        end
+
+        it 'tries hold lookup again until item info returns not empty' do
```
cdmo: It isn't clear to me how this is being tested, sorry. How does this spec prove that a second call to the symphony client is made when items is `nil`? I think I was expecting some kind of (pseudo code) `SymphonyClient calls.count == 2` or something, idk. This seems like it's just showing that the second object passed back from the `SymphonyClientParser` is valid. Sorry again if I'm totally missing this, it's been a while since I've looked at this.

spec/controllers/holds_controller_spec.rb
banukutlu: No, it will return one object for each call, but first call will return the first arg, and the second will return the second one.

spec/services/symphony_client_spec.rb
```diff
@@ -203,6 +216,17 @@
...
       end
     end
+
+    context 'when place hold return records in use error' do
+      let(:item_barcode) { 'records_in_use_barcode' }
+      let(:error_prompt) { { messageList: [{ code: 'hatErrorResponse.116', message: 'Records in use' }] }.to_json }
+
+      it 'keeps trying for 5 seconds max until the record is not in use' do
+        place_hold_response = client.place_hold(patron, user.session_token, item_barcode, hold_args)
+
```
cdmo: Also here, I wonder if we can make the spec actually prove that the loop is running. Maybe like an expectation that `sleep` gets invoked? 🤔

spec/controllers/holds_controller_spec.rb
banukutlu: https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/returning-a-value#specify-different-return-values-for-multiple-calls

cdmo: so SymphonyClientParser.parsed_response is called twice?

cdmo: wow, :the_more_you_know:

I don't think I've seen `and_return` used like this before, nice!

spec/controllers/holds_controller_spec.rb
banukutlu:
```
allow(SymphonyClientParser).to receive(:parsed_response).with(mock_client, :get_hold_info, *hold_info_params)
            .and_return(parsed_hold_info_empty_item, parsed_hold_info)
```
First call to SymphonyClientParser returns the empty item response and the second call returns the response with hold info. So, I did not think we necessarily have to test the empty case or the number of totals calls as long as the final result is the response with hold info. But I think we can add another example to test the calls count if you think it makes it more clear.

cdmo: No we're good, I got it now from our conversation earlier in the issue, thanks for spelling it out 👍

spec/services/symphony_client_spec.rb
banukutlu: I am looking into this

app/services/symphony_client.rb
banukutlu: Might even be added to PlaceHoldResults::Builder 🤔 (which is added in the next PR)

spec/services/symphony_client_spec.rb
cdmo: explained it to me in slack, yes this is the same as the other stubbed request and_return thing with multiple responses 👍

Issue #112, opened 2020-02-11
Group things.


cdmo:
Stems from #106 

* Group all non-volumetrics (see `476022`, ~the most recent publication~ is grouped across locations and `15513` - actually this isn't the "most recent publication" - it's actually more likely to be "things we don't understand better" but we are still grouping them together because we lack better data)
* Group volumetrics of the same call together (see `7517385`)

This will mimic e-library

cdmo:
After discussing with @mkutch and Jeff in Maryam's office, we went over some scenarios and believe that it may be best for user experience to show call numbers always and to not group non-volumetrics. Tagging this as question so we can bring it up at our next meeting.

cdmo:
We talked about this during our check in meeting, indeed we will not be grouping non-volumetrics.

cdmo:
After discussing with @mkutch and Jeff Edmunds, we've landed at "we should indeed continue to group non-volumetric holdables" - the reason being that there is a bit of a black box in terms of how a hold placement is processed once received and we want to play it safe and mirror the current process so as not to disrupt service/process.

cdmo:
volumetrics that are of different call must be grouped in the new request form too.

non-volumetrics are grouped regardless of call number.

cdmo:
@ruthtillman We group all non-volumetrics together regardless of call. So this can get a little weird. For example, see `15513`. In e-library we don't display call so it's just a parenthetical with locations (which is not good UX of course). In MyAccount we are display call number of course. So, I think we can just not display a call number when grouped? Or display all of them? Not sure.

![Screenshot 2020-03-03 15 43 00](https://user-images.githubusercontent.com/523381/75817720-acc6a100-5d65-11ea-887b-cf49123b1e25.png)

cdmo:
Maybe it should just be like:

* [ ] Book

So just displaying the item type.

cdmo:
For now I'm just picking the first Call and using it's call number. So in the above case that's 

![Screenshot 2020-03-04 15 59 14](https://user-images.githubusercontent.com/523381/75922501-1c559280-5e31-11ea-83e8-8b49544419ce.png)

ruthtillman:
Check what's happening with Shenango in http://myaccount-qa.herokuapp.com/holds/new?catkey=7517385

ruthtillman:
[...]

PR #150, opened 2020-03-04
112 group things

cdmo:
Issue #112 

calling this "Group things" was a mistake, really it's about "flattening." 

Additional change: updated rubocop settings

### DIFF
--- a/app/services/place_hold_form/builder.rb
+++ b/app/services/place_hold_form/builder.rb
@@ -11,7 +11,7 @@ class PlaceHoldForm::Builder
 
   def generate
     bib_info = Bib.new(parsed_symphony_response(:get_bib_info, @catkey, @user_token))
-    @call_list = bib_info.call_list.map { |c| Call.new record: c }
+    @call_list = bib_info.call_list.map { |call| Call.new record: call }
     process_volumetric_calls
 
     {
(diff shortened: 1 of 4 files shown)


app/services/place_hold_form/builder.rb
```diff
@@ -44,10 +44,24 @@ def parsed_symphony_response(symphony_call, *params)
...
       @volumetric_calls = @call_list.dup if volumetric? && @call_list.count > 1
       volumetric_natural_sort
+      compact_non_volumetric_calls if @volumetric_calls.select { |c| c.volumetric.nil? }.count > 1
+      @volumetric_calls.uniq!(&:call_number)
+    end
+
+    def compact_non_volumetric_calls
+      process_these = @volumetric_calls.dup
+      @volumetric_calls = []
+      process_these.each_with_index do |c, i|
```
banukutlu: a minor thing but per our conversation, renaming `c`  to be more descriptive would be helpful here

cdmo: rename `c` or `process_these`?

i use `c` through the class as shorthand for `call`... 🤔 

`process_these` was a var i just forgot to rename before PRing... maybe I should rename both?

banukutlu: oh I think `process_these` is fine, sorry I just meant `c`. I mean if you want to rename the `process_these` as well, your call but it is clear to read to me as is :)

cdmo: okie doke, hope that improves readability

banukutlu: looks great, thanks for the updates!

PR #152, opened 2020-03-06
redis for session (and cache)

cdmo:
a little too easy to setup. More to come soon

Relevant:
`activesupport-6.0.1/lib/active_support/cache/redis_cache_store.rb`

## Additional changes

Added a "ubiquitous language" section (i.e., terminology)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -20,7 +20,6 @@ Rails.application.configure do
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :memory_store
     config.public_file_server.headers = {
       'Cache-Control' => "public, max-age=#{2.days.to_i}"
     }
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,11 @@ gem 'turbolinks', '~> 5'
 gem 'warden', '~> 1.2'
 gem 'webpacker', '~> 4.0'
 
+group :production do
+  gem 'hiredis'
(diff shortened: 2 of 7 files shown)


config/environments/development.rb
```diff
@@ -20,7 +20,6 @@
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :memory_store
```
cdmo: Not sure about the extra cache settings in this block, like whether they are necessary or not. For now, not touching the, but memory_store has to be removed otherwise it would override the setting in application.rb.

cdmo:
deploying to heroku

cdmo:
Left on this issue: try in heroku

If that passes, get review and input for production system.

cdmo:
though we still have production setup to figure out, we can do that later as a separate PR I'm thinking (it'll be a matter of juggling a new env var and nothing else 🤞). In the meantime this can be reviewed and merged if passing.

cdmo:
thanks for the help @whereismyjetpack, feel free to chime in if you like.

whereismyjetpack:
@cdmo the only thing that looks off kilter is how a redis with password auth would work. the redis driver _should_ take a password: hash object, but it's not clear if that will work in this scenario. if it does, we should send it. (in redis clients that i've worked with, sending `nil` is as good as not setting it, so you could in your configuration set `redis_password = ENV.fetch('REDIS_PASSWORD', nil)`

Issue #151, opened 2020-03-05
Refactor place hold results


banukutlu:
move from holds controller to a service

PR #154, opened 2020-03-06
Refactor place hold results

banukutlu:
Issue #151 

## Note
✔ #147 needs to be merged before this

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -72,18 +72,14 @@ class HoldsController < ApplicationController
     return redirect_to holds_path if session[:place_hold_results].blank?
 
     @place_hold_catkey = session[:place_hold_catkey]
+
+    results_builder = PlaceHoldResults::Builder.new(user_token: current_user.session_token,
+                                                    client: symphony_client,
+                                                    place_hold_results: session[:place_hold_results])
+    @place_hold_results = results_builder.generate
+
     session.delete(:place_hold_catkey)
-
-    raw_place_hold_results = session[:place_hold_results]
     session.delete(:place_hold_results)
-
-    @place_hold_results = raw_place_hold_results.each do |status, results|
-      if status == 'success'
-        results.map { |result| result['placed_hold'] = hold_lookup(result['hold_key']) }
-      else
-        results.map { |result| result['failed_hold'] = item_lookup(result['barcode']) }
-      end
-    end
   end
 
   # Handles form submission for canceling holds in Symphony
@@ -164,26 +160,6 @@ class HoldsController < ApplicationController
       flash[:error] = t 'myaccount.hold.update_pickup.past_date'
 
       redirect_to holds_path
-    end
-
-    # Fetch a hold lookup as many times as necessary up to 5 seconds to avoid `nil` item in response.
-    def hold_lookup(hold_key)
-      parsed_hold = SymphonyClientParser::parsed_response(symphony_client,
-                                                          :get_hold_info, hold_key, current_user.session_token)
-
-      start = DateTime.now
-      while parsed_hold.dig('fields', 'item').nil? && start + 5.seconds > DateTime.now
-        parsed_hold = SymphonyClientParser::parsed_response(symphony_client,
-                                                            :get_hold_info, hold_key, current_user.session_token)
-      end
-
-      Hold.new parsed_hold
-    end
-
-    def item_lookup(barcode)
-      parsed_item = SymphonyClientParser::parsed_response(symphony_client,
-                                                          :get_item_info, barcode, current_user.session_token)
-      Item.new parsed_item
     end
 
     def check_for_blanks!
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -166,6 +166,8 @@ class SymphonyClient
       return if response.status.ok?
 
       JSON.parse(response.body).dig('messageList')[0].dig('code')
+    rescue JSON::ParserError
+      nil
     end
 
     def records_in_use?(response)
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -165,7 +165,7 @@ class SymphonyClient
     def error_code(response)
       return if response.status.ok?
 
-      JSON.parse(response.body).dig('messageList')[0].dig('code')
+      JSON.parse(response.body).dig('messageList').first.dig('code')
     rescue JSON::ParserError
       nil
     end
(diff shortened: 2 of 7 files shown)


app/services/place_hold_form/builder.rb
```diff
@@ -10,7 +10,7 @@ def initialize(catkey:, user_token:, client:)
   end
 
   def generate
-    bib_info = Bib.new(parsed_symphony_response(:get_bib_info, @catkey, @user_token))
+    bib_info = Bib.new(SymphonyClientParser::parsed_response(@client, :get_bib_info, @catkey, @user_token))
```
cdmo: Thanks for updating this

app/controllers/holds_controller.rb
```diff
@@ -72,18 +72,14 @@ def result
...
-    session.delete(:place_hold_catkey)
 
-    raw_place_hold_results = session[:place_hold_results]
-    session.delete(:place_hold_results)
+    results_builder = PlaceHoldResults::Builder.new(user_token: current_user.session_token,
+                                                    client: symphony_client,
+                                                    place_hold_results: session[:place_hold_results])
+    @place_hold_results = results_builder.generate
 
-    @place_hold_results = raw_place_hold_results.each do |status, results|
```
cdmo: really happy to see this code and related methods moved out of the controller

app/services/symphony_client.rb
```diff
@@ -166,6 +166,8 @@ def error_code(response)
       return if response.status.ok?
 
       JSON.parse(response.body).dig('messageList')[0].dig('code')
+    rescue JSON::ParserError
+      nil
```
cdmo: I want to make sure I understand this bit: When JSON tries to parse the response body and then dig for code, if it hits a bad formatting problem or whatever in the JSON itself, it will raise up and set error_code to nil?

app/services/symphony_client.rb
```diff
@@ -166,6 +166,8 @@ def error_code(response)
       return if response.status.ok?
 
       JSON.parse(response.body).dig('messageList')[0].dig('code')
```
cdmo: Not part of this PR, but what do you think of:

```suggestion
      JSON.parse(response.body).dig('messageList').first.dig('code')
```

I think it's easier to read, more ruby-ish

app/services/symphony_client.rb
banukutlu: yup, will return nil for any response without the expected messageList and code fields

cdmo: this would only catch errors during `parse` not during `dig` or `first`... right?

spec/services/place_hold_results/builder_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
```
cdmo: Should we add a spec to test that the loop is being executed too? Like you did for #147?

spec/services/place_hold_results/builder_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  describe '#generate' do
+    it 'will pass along processed placed holds results' do
+      expect(processed_results[:success].first).to include(hold_key: 'a_hold_key')
+        .and include(barcode: 'holdable_barcode')
+        .and include(:placed_hold)
+    end
+
+    it 'returns Hold objects for placed holds' do
+      expect(processed_results[:success].first[:placed_hold]).to be_a Hold
```
cdmo: This spec and the one that starts on 49 use type checking. I kinda think it's good to avoid type checking, and maybe instead check for something unique about Hold. For example:

`expect(processed_results[:success].first[:placed_hold].pickup_library_human).to eq 'Pattee Library and Paterno Library Stacks'`

I think this makes the code more flexible, less brittle. If we ever renamed `Hold` this spec would start failing as it is, but if we just test against a unique characteristic of Hold (like the human pickup library name) it wouldn't.

For `Item` we could use `barcode` for example.

What do you think?

app/services/symphony_client.rb
banukutlu: I think so yes

cdmo: I wonder if it'd `raise` this if there was a problem, like if dig or first hit a fatal. Anyway, this is good.

spec/services/place_hold_results/builder_spec.rb
banukutlu: if you mean the records in use workaround, I am thinking that we test it with symphony_client_spec, and results just makes calls to symphony_client methods so kinda leaning towards to no need to that here again but would like to hear your reasoning, I might be missing sth.

spec/services/place_hold_results/builder_spec.rb
banukutlu: I see your point, just trying to find uniqueness for Item 🤔 I am not sure if barcode is unique to Item only.

spec/services/place_hold_results/builder_spec.rb
cdmo: So the workaround in #147 relies on `SymphonyClient::authenticated_request` to delay if `records_in_use?`. This one is different. It is checking for a malformed response from the WS. Then it will delay here, in this class (not in SymphonyClient::autheticated_request`. We are not testing a scenario where this results builder encounters malformed response to looking up a hold. 

Strange to me that CC is saying that the line inside the `while` loops is covered, so maybe I'm mis-reading.

spec/services/place_hold_results/builder_spec.rb
banukutlu: ok, makes sense, I changed the Hold spec as you suggested and deleted the Item one since there is already another example checking the barcode. So, it should be good now

spec/services/place_hold_results/builder_spec.rb
banukutlu: ahhhh ok yeah, the spec for that is still at holds_controller_spec https://github.com/psu-libraries/myaccount/blob/151-refactor-results/spec/controllers/holds_controller_spec.rb#L367, I will try to move it over

spec/services/place_hold_results/builder_spec.rb
cdmo: Yup. I think this is good too. Thanks.

spec/services/place_hold_results/builder_spec.rb
cdmo: Ah that's why some confusion. Hope it's an easy port over. Thanks!

banukutlu: yes sorry, I apparently forgot to update these, working on it now :)

cdmo: No worries!!

banukutlu: ok, pushed a commit for this change, hope it looks right

PR #157, opened 2020-03-13
place hold fix

banukutlu:
Issue #96

## merge first (in order):
- [x] #147 
- [x] #154 
- [x] #164

### DIFF
--- a/app/services/place_hold_results/builder.rb
+++ b/app/services/place_hold_results/builder.rb
@@ -10,7 +10,7 @@ class PlaceHoldResults::Builder
 
   def generate
     @place_hold_results.each do |status, results|
-      if status == 'success'.to_sym
+      if status == :success
         results.map { |result| result[:placed_hold] = hold_lookup(result[:hold_key]) }
       else
         results.map { |result| result[:failed_hold] = item_lookup(result[:barcode]) }
--- a/app/services/place_hold_results/builder.rb
+++ b/app/services/place_hold_results/builder.rb
@@ -10,10 +10,10 @@ class PlaceHoldResults::Builder
 
   def generate
     @place_hold_results.each do |status, results|
-      if status == 'success'
-        results.map { |result| result['placed_hold'] = hold_lookup(result['hold_key']) }
+      if status == 'success'.to_sym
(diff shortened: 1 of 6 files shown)


cdmo:
I think it's okay for this to be merged independently of #154 right? This branch does not build on top of the branch in #154 (ie `151-refactor-results`). actually I think #147 is separate too. Trying to understand why those have to go first 🤔

cdmo:
this one *is* dependent on #164 getting merged first though. I made that change because it's easier on my eyes to keep these two changes separate, but this place hold fix does build on top of the sessions fix, so that one should get merged in first to avoid a messy situation..

cdmo:
Sorry, deleted my earlier comments, yes, #154 should be merged first to avoid messy merge. Then #164.

app/services/place_hold_results/builder.rb
```diff
@@ -10,10 +10,10 @@ def initialize(user_token:, client:, place_hold_results:)
 
   def generate
     @place_hold_results.each do |status, results|
-      if status == 'success'
-        results.map { |result| result['placed_hold'] = hold_lookup(result['hold_key']) }
+      if status == 'success'.to_sym
```
cdmo:
```suggestion
      if status == :success
```

banukutlu: > One minor change, but my bigger question is - what changed that made the access not-so-indifferent anymore? Why are symbols working where strings once were? This keys off of `HoldsController::create` where the keys of the array are `:success` and `:error` and have been for a while 🤔

yup, I had the same questions myself and honestly do not know the answer, did not have much time to dig into it :/

banukutlu:
> One minor change, but my bigger question is - what changed that made the access not-so-indifferent anymore? Why are symbols working where strings once were? This keys off of `HoldsController::create` where the keys of the array are `:success` and `:error` and have been for a while 🤔

yup, I had the same questions myself and honestly do not know the answer, did not have much time to dig into it :/

cdmo:
Ok. Well, I think maybe it shouldn't have worked with indiferrent access unless somewhere along the line we added that method somewhere. Maybe we got rid of it when you cleaned up the JSON fixtures. Anyway, the way it is now is how it should be.

Issue #162, opened 2020-03-17
Don't link if shadowed *OR if it's equip7day5*


cdmo:
See holobar's account, it's trying to link to a shadowed item

Thanks @mkutch for finding this

@mkutch can you let us know what field indicates shadowed?

mkutch:
@cdmo When you do a bib level web service call the field is called `shadowed` and the values are true and false.

cdmo:
Perfect, thanks.

ruthtillman:
Checking Holobar's account, the Course Reserve isn't showing up but equip7day5 test up (catkey 29014105) is still showing up linked... is this a different to-do?

banukutlu:
> Checking Holobar's account, the Course Reserve isn't showing up but equip7day5 test up (catkey 29014105) is still showing up linked... is this a different to-do?

That one has an item type that was only defined in test afaik, right @mkutch. We do not link several item types and this should be one of those normally.

mkutch:
@banukutlu Yes, that item type is not in production policy.

cdmo:
that will be fixed with #130

cdmo:
#130 will provide dynamic itemType translations, and I can also add the filter for equip7day5 there too. Sound good?

banukutlu:
> #130 will provide dynamic itemType translations, and I can also add the filter for equip7day5 there too. Sound good?

I think there is no need to add this type, afair that one is only added for testing purposes. @mkutch?

cdmo:
Thought it was in test because it was going to be added to prod eventually?

mkutch:
@banukutlu I emailed Chris asking if equipp7day5 was going to move into prod. His response

> I guess it was on my list to move over before the world ended

cdmo:
So i'll add it to our exclusion list of linked items. (just coordinating with @banukutlu so we don't both do it)

PR #163, opened 2020-03-20
unlink shadowed items

banukutlu:
Issue #162


app/services/symphony_client.rb
```diff
@@ -7,6 +7,11 @@ class SymphonyClient
     content_type: 'application/json'
   }.freeze
 
+  ITEM_RESOURCES = [
+    'bib{title,author,callList{*}}',
+    'item{*,bib{shadowed,title,author},call{sortCallNumber,dispCallNumber}}'
+  ].join(',')
+
```
cdmo: Moving this up with the rest of the constants makes sense I think. Why'd you turn it into an array with two values (bib and item,call)? I think either it should stay a simple string or maybe break it into bib, item and call (3 values of the array) to improve readability. Assuming this change had to be made due to line length?

cdmo: The only other thing I am struggling to wrap my head around with all of this now is why we are calling for title and author from `bib` and `item>bib`  and then (relatedly) pulling `shadowed` from `item>bib` (I know below for hold lookup and item lookup both are pulling `shadowed` from just the bib level).

Had to look again at checkouts and holds.

For checkouts, the `circRecordList`'s `circRecord`s do not contain `bib` (even though we ask for it as part of the web service request). It'll pass bib info along because we specified it in the call (i.e., `item{*,bib{shadowed,title,author}`).

For holds, the `holdRecordList`'s `holdRecord` _does_ contain `bib`. (aside, we could drop the ask for title and author from `bib` or `item>bib` on the holds page - right now I guess it's a bit redundant, but probably not a problem in terms of speed).

I didn't look at fines because the test account I had didn't have any. But it doesn't matter really because I see why we can't just always use "top-level" `bib`. Sometimes `bib` is available at the "top-level" and sometimes it isn't (like for `circRecord`) - *but* `bib` is always available at the `item > bib` level.

Hope this rambling made some sense, just had to walk back through the symphony request logic. No change asked for, to be clear.

cdmo:
oh guess it has to be rebased

app/services/symphony_client.rb
banukutlu: > Moving this up with the rest of the constants makes sense I think. Assuming this change had to be made due to line length?

yes, because of the line length rubocop was getting annoyed

banukutlu: > 
> I didn't look at fines because the test account I had didn't have any. But it doesn't matter really because I see why we can't just always use "top-level" `bib`. Sometimes `bib` is available at the "top-level" and sometimes it isn't (like for `circRecord`) - _but_ `bib` is always available at the `item > bib` level.
> 
> Hope this rambling made some sense, just had to walk back through the symphony request logic. No change asked for, to be clear.

yes, sometimes it is not available at the top-level and that may be redundant really, let's  chat about this to see if we can remove that part.

cdmo: Does sorta make me wonder if the dynamic loading of arguments to the web service is a little _too_ fancy and complicated. But, again, no change needed.

PR #164, opened 2020-03-21
Sessions

cdmo:
I chopped off the last commit added to #157. Just to make it easier for me to review. Hope that's ok.


app/controllers/application_controller.rb
```diff
@@ -41,7 +35,17 @@ def patron_info_response
     end
 
     def authenticate_user!
-      redirect_to root_url unless current_user?
+      if current_user?
+        return renew_session_token if patron.stale?
+      else
+        redirect_to root_url
+      end
```
cdmo: For me this is a little more readable, just adding one line (`renew_session_token if patron.stale?
`) instead of a kind of complex nested if thing:

```suggestion
      redirect_to root_url unless current_user?
      renew_session_token if patron.stale?
```

banukutlu:
> @banukutlu let's just delete `app/views/sessions/index.html.erb` and `spec/views/sessions/index.html.erb_spec.rb`. We don't need them. Right?

yes, I thought about that too but was not sure if we ever need them, wanted to see what you will say, so yes for deleting ✔️

banukutlu:
> @banukutlu I'm curious to know if you've tested this in the wild. Like, have you let the application site with an authenticated user so long that it gets stale? Does the fix work?

yeah, I tried testing it with letting it to get stale, it seemed to work but since it is not easy to test, I hope these changes were actually covering it properly 🤔

cdmo:
They look right to me, glad you tested. This is ready now, right? or do you have more changes?

app/controllers/application_controller.rb
```diff
@@ -36,7 +36,7 @@ def patron_info_response
 
     def authenticate_user!
       redirect_to root_url unless current_user?
-      renew_session_token if patron.stale?
+      renew_session_token if current_user? && patron.stale?
```
cdmo: 🤔 so, curious why you added a check for `current_user?`? Doesn't the `unless` above mean that `current_user?` resolves to `true`? Is this redundant? 

Also, your commit says "

"since redirect_to does not return, this line will be executed so check if current_user, otherwise errors out with undefined patron." Here is my understanding: There are 3 scenarios:

1) `current_user?` is false - redirect this to the root_url (which will start up authentication)
2) `current_user?` is true (which is implied in the world that exists after the first line of the method):

In 2) either the patron.stale? is true or false. If it is true, then the renew_session_token method is called. If it's false, then the method has basically done nothing and returns `nil`.

Is that wrong? I'm not sure I see where erroring out with an undefined patron happens?

banukutlu: we chat about this and decided that we will use `return redirect_to` since we both like that way better.

cdmo: Just for future me if I'm re-reading, without `return` the second line in the method gets processed. Thanks @banukutlu

cdmo:
Ok @banukutlu this is ready to merge. I can't approve it, you'll have to do that. It's like that because I split one of your PRs into two. Thanks. Please approve and merge in

Issue #161, opened 2020-03-17
Make item type display and translation dynamic


cdmo:
Leverage caching too

See holobar's account, there's a new item type that has no translation. (`EQUIP7DAY5` or something)

Thanks @mkutch for finding this

cdmo:
Just want to mention that the only thing that needs review here is that a) the item type descriptions still show up correct and b) that the new `equip7day5` is unlinked

PR #165, opened 2020-03-25
161 dynamic item type mapping

cdmo:
#161 

@banukutlu can you look at what I've done here? Basically the item type lookup is only done once per request. This saves the server from having to go out X times for this static data in a single request (X being the number of bibrecords like holds or checkouts). There are _definitely_ other ways to accomplish this. Would like your initial feedback before I write specs. Is this approach ok? Would you rather do this another way?

I did think about storing this array response in a session variable, but session is not available at the model level and it felt wrong anyway to do that in a model. So then it'd have to be done in... maybe a service object or stray controller that gets included by every bibrecord related controller (i.e., Holds, Fines, Checkouts)? Just seemed confusing to me.

Also thought about caching this value in some global store in Redis. That could work too, I'm just not familiar with the how. 

In the end this approach is to me pretty straightforward and relatively inexpensive.

Let me know what you think.

*Additional change*: Updated README.md for fix on redis docker stuff. And #162


cdmo:
Oh also initially I used class variables for this, but apparently they are frowned upon (by rubocop and others) due to the confusion they can cause during inheritance. See https://www.ruby-lang.org/en/documentation/faq/8/

cdmo:
@banukutlu I'm inclined to ignore CC's warning on this. Alternatively we could just leave the "base" methods in SC like request and authenticated request and then separate everything else based on grouping? SCHolds, SCLogin, SCPatron... idk. The file is getting ungainly, but I don't have a better solution that wouldn't involve a fair amount of work to refactor.

cdmo:
Also note that we have a lot of failing specs, this is because I'd have to now stub a ws request for anywhere item type is used.

banukutlu:
> then

agreed :100:

banukutlu:
> #161
> 
> @banukutlu can you look at what I've done here? Basically the item type lookup is only done once per request. This saves the server from having to go out X times for this static data in a single request (X being the number of bibrecords like holds or checkouts). There are _definitely_ other ways to accomplish this. Would like your initial feedback before I write specs. Is this approach ok? Would you rather do this another way?
> 
> I did think about storing this array response in a session variable, but session is not available at the model level and it felt wrong anyway to do that in a model. So then it'd have to be done in... maybe a service object or stray controller that gets included by every bibrecord related controller (i.e., Holds, Fines, Checkouts)? Just seemed confusing to me.
> 
> Also thought about caching this value in some global store in Redis. That could work too, I'm just not familiar with the how.
> 
> In the end this approach is to me pretty straightforward and relatively inexpensive.
> 
> Let me know what you think.
> 
> _Additional change_: Updated README.md for fix on redis docker stuff.

I think your approach looks good ✔️ The only other idea I like is to cache the values in a global store in Redis, but as you mentioned current approach seems more straightforward for now. So, yeah let's go with this :) Thanks so much 👍

cdmo:
Thanks!

I'm a little more familiar with "the how" now after playing with redis (and sidekiq) for the past few days. Might actually be a better idea to store this in a hash in redis and instruct rails to invalidate that key where it lives every 24 hours. Redis is slick.

banukutlu:
> Thanks!
> 
> I'm a little more familiar with "the how" now after playing with redis (and sidekiq) for the past few days. Might actually be a better idea to store this in a hash in redis and instruct rails to invalidate that key where it lives every 24 hours. Redis is slick.

I am all in favor for redis option 👌 and yes caching for 24hrs sounds good.

cdmo:
I'll just update that get mapping method so that it is getting and setting from Redis then? I think that should be fine.

cdmo:
this is ready now

cdmo:
pending again, adding #162

cdmo:
ready again

app/models/concerns/bib_record.rb
```diff
@@ -103,4 +59,14 @@ def bib
...
     end
+
+    # Use Redis to store this globally once an hour
+    def item_type_mapping
+      redis = Redis.new
+      return JSON.parse(redis.get('item_type_map')) unless redis.get('item_type_map').nil?
+
+      client = SymphonyClient.new
+      redis.setex('item_type_map', 3600, client.get_item_type_map.to_json)
+      JSON.parse(redis.get('item_type_map'))
```
banukutlu, line 70:
```suggestion
      item_type_map = client.get_item_type_map.to_json
      redis.setex('item_type_map', 3600, item_type_map)
      JSON.parse(item_type_map)
```

@cdmo with not being so sure but would the above be better to avoid `redis.get` here? maybe it is an insignificant change, please let me know what you think?

cdmo: Yes thanks, good spot!

spec/models/item_spec.rb
```diff
@@ -5,6 +5,18 @@
 RSpec.describe Item, type: :model do
   subject(:item) { build(:item) }
 
+  let(:redis_instance) { instance_double(Redis) }
+  # let(:symphony_client_instance) { instance_double(SymphonyClient) }
```
banukutlu, line 9:
```suggestion
```

cdmo: oops forgot to delete, thanks :)

cdmo:
Just gonna leave this article on dependency injection here in case we want to give it a shot some day (yes I realize this is my PR 😆 ). Still feel kinda weird about leaving the class dependencies in a private method kinda buried, but, I think it's okay.

https://medium.com/@Bakku1505/introduction-to-dependency-injection-in-ruby-dc238655a278
https://github.com/dry-rb/dry-container
https://github.com/dry-rb/dry-auto_inject

Issue #153, opened 2020-03-06
Investigate async-ing symphony client calls


cdmo:
Redis, sidekiq, active job, and [Bootstrap Toasts](https://getbootstrap.com/docs/4.3/components/toasts/) potentially

cdmo:
POC is in a working state at #166. Uses sidekiq and redis to turn the application into a partial API. This is one option. POC is only working on pickup library location changes for now.

cdmo:
I'm going to mark this as closed and future implementation work (rather than this investigation issue) will be split out into separate issues.

cdmo:
General strategy outlined here:

https://github.com/psu-libraries/myaccount/pull/166#issuecomment-611670938

PR #166, opened 2020-03-27
Adds Sidekiq and Redis along with polling functionality for pickup change location async

cdmo:
…y POC #153

## UPDATE:

Now linked to #173


cdmo:
[before.log](https://github.com/psu-libraries/myaccount/files/4391167/before.log)  
[after.log](https://github.com/psu-libraries/myaccount/files/4391165/after.log)  
[after_with_redirect.log](https://github.com/psu-libraries/myaccount/files/4391166/after_with_redirect.log)

queuing jobs to happen later still took 11 seconds.

cdmo:
ok, so the extra 11 seconds came from an unnecessary call to `ApplicationController::patron`. 11 seconds makes sense for a patron with dozens and dozens of items. This patron call occurs merely to check if patron is stale before sending request to do the batch update. It fires via befoe action `authenticate_user!`. If you take out the check on patron.stale? it speeds up significantly. So, after removing that stale check: 

updating the pickup libraries action on 48 holds:
* with sidekiq: nearly instantaneous at .179s
* without sidekiq: ~9.5s

This is better than 20 seconds, but still not great.

So, maybe, even if we don't end up using sidekiq for processing updates/deletes, or even _reads_ (!) is that we change the way we check for stale. Maybe we make a separate stale_check action and run it before index methods on holds, fines, and checkouts?

I do think using sidekiq will force some pretty significant changes. Changes involving javascript and maybe even ActionCable. But basically we'd have to replace redirects to index pages with in-place notifications and updates.

banukutlu:
> [before.log](https://github.com/psu-libraries/myaccount/files/4391167/before.log)
> [after.log](https://github.com/psu-libraries/myaccount/files/4391165/after.log)
> [after_with_redirect.log](https://github.com/psu-libraries/myaccount/files/4391166/after_with_redirect.log)
> 
> queuing jobs to happen later still took 11 seconds.

still better than before 👍

banukutlu:
>  If you take out the check on patron.stale? it speeds up significantly. So, after removing that stale check:
> 
> * with sidekiq: nearly instantaneous at .179s
> * without sidekiq: ~9.5s
> 
> This is better than 20 seconds, but still not great.

yeah, we should def. do sth about the patron.stale? check

cdmo:
wondering if we shouldn't just set a timer. patron.stale? is so expensive when the action isn't index for folks with a lot of records. Maybe we authenticate and record the time... set it in redis... and then only reauth when an hour has passed? Maybe we could find out what the definitive session length is and make the timer even longer?

cdmo:
latest commit is a total mess, but the application is now in a state that can be called POC for using sidekiq and redis plus async/await javascript to update pickup library. so this is one option.

async

cdmo:
marked ready for review, but definitely not ok to merge, more looking for comments on approach. plus the session stuff needs sorting out first.

cdmo:
oh meant to share some sites I looked at:

Sidekiq and async

* https://www.youtube.com/watch?v=CStZg8ql9Vs&list=UUxJaNyXCQw0mghY0hA1wA9w&index=159
* redis set noeviction (aside: can't remember why i wrote this)
* https://edgeguides.rubyonrails.org/active_job_basics.html
* sidekiq monitoring https://github.com/mperham/sidekiq/wiki/Monitoring
* https://charusat09.github.io/update_frontend_using_sidekiq_worker/
* https://medium.com/@AdamKing0126/ajax-and-rails-demystifying-remote-true-fe51ba2ce819 (aside: it's still mystifying)

app/javascript/packs/hello_erb.js.erb
```diff
@@ -0,0 +1,6 @@
+// Run this example by adding <%%= javascript_pack_tag 'hello_erb' %> to the head of your layout file,
```
banukutlu, line 1: @cdmo can we delete this file?

cdmo: yeah, and in fact a lot of this PR can be deleted, it's a total mess! 🙃

cdmo:
I looked into StimulusJS and wasn't overly thrilled after all. Basically it does what we already do which is extensively using and depending on data- attributes. The main thing it gives you are some convenience methods around setting instance variables and probably more importantly, a more consistent design of our JS scripts. Beyond that not a lot.

One thing we talked about today as a pro of sidekiq is retries, and it could potentially remove some of our retry code in other classes.

Something else I was thinking of is for handling very long running tasks, tasks over 30 seconds. Because even in Javascript those calls will fail due to a timeout. With sidekiq they won't. That's a nice safety in general, and kind of a necessity for _some_ heavy users.

cdmo:
Just some thinking outloud:

* We'll definitely be doing async
* We'll probably use sidekiq (with redis) because of long-running jobs and wanting to avoid timeouts (even in async js calls they'd 500) - this is to say i don't think we can do this without sidekiq. Sidekiq also gives the added [benefit of exponential back off retries](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry). The cost of sidekiq is the overhead of running another tool and the complexity of configuring and deploying it.
* We probably don't need to "asyncify" all the SymphonyClient calls. Like logging in and your new stale check. The profile information calls are also very efficient. I think we can get away with not asyncifying fines, but we probably should just in case there are some very heavily fined patrons out there.
* We'll have to asyncify the display of holds, checkouts/renewals.
* I'm thinking the above will look like this: 

1. open up /holds
1. holds#index runs (checks session freshness, but doesn't call symphony WS server-side) 
1. page is rendered with the chrome of the site and in the middle there is a message like "We're retrieving your items now" or maybe just a giant spinner or maybe those fancy lines (ahem ["skeleton css"](https://css-tricks.com/building-skeleton-screens-css-custom-properties/)) 
1. javascript function then triggers a sidekiq job to get holds (will have to build out a controller for processing sidekiq jobs - this PR has delete and get only, no create) 
1. JS function using async and await patiently tries again and again to see if the job is done. Once it is done, the response can be retrieved from redis (where the sidekiq job will store the data)

* We'll also have to asyncify edits that happen to holds and checkouts and we can follow a pattern similar to this PR
* I think we might be able to create a more generic functions in JS that can be shared among all these calls

Just wanted to get some of these thoughts out.

cdmo:
Just some thinking outloud:

* We'll definitely be doing async
* We'll probably use sidekiq (with redis) because of long-running jobs and wanting to avoid timeouts (even in async js calls they'd 500 for users with lots of stuff, like a 500 item checkouts request takes 45 seconds) - this is to say i don't think we can do this without sidekiq. Sidekiq also gives the added [benefit of exponential back off retries](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry). The cost of sidekiq is the overhead of running another tool and the complexity of configuring and deploying it.
* We probably don't need to "asyncify" all the SymphonyClient calls. Like logging in and your new stale check. The profile information calls are also very efficient. I think we can get away with not asyncifying fines, but we probably should just in case there are some very heavily fined patrons out there.
* We'll have to asyncify the display of holds, checkouts/renewals.
* I'm thinking the above will look like this: 

1. open up /holds
1. holds#index runs (checks session freshness, but doesn't call symphony WS server-side) 
1. page is rendered with the chrome of the site and in the middle there is a message like "We're retrieving your items now" or maybe just a giant spinner or maybe those fancy lines (ahem ["skeleton css"](https://css-tricks.com/building-skeleton-screens-css-custom-properties/)) 
1. javascript function then triggers a sidekiq job to get holds (we need to think about this some more) 
1. JS function using async and await patiently tries again and again to see if the job is done. Once it is done, the response can be retrieved from redis (where the sidekiq job will store the data)

* We'll also have to asyncify edits that happen to holds and checkouts and we can follow a pattern similar to this PR
* I think we might be able to create more generic functions in JS that can be shared among all these calls

Just wanted to get some of these thoughts out.

app/controllers/holds_controller.rb
```diff
@@ -116,19 +116,11 @@ def holds_not_ready
...
-      when 200
-        process_flash(:success, 'update_pickup.success_html')
-      else
-        Rails.logger.error(change_pickup_response.body)
-        process_flash(:error, 'update_pickup.error_html')
-      end
+    def handle_pickup_change_request(pass_these)
+      # symphony_client = SymphonyClient.new
+      # symphony_client.change_pickup_library(hold_key: pass_these[:hold_key], pickup_library: pass_these[:pickup_library], session_token: pass_these[:session_token])
+
```
banukutlu, line 122:
```suggestion
```

app/controllers/holds_controller.rb
```diff
@@ -18,14 +18,14 @@ def index
...
-  def update
-    params['hold_list'].each do |holdkey|
-      @hold_to_act_on = holds.find { |hold| hold.key == holdkey }
-      handle_pickup_change_request if params['pickup_library'].present? && params['pickup_library'].present?
-      handle_not_needed_after_request if params['pickup_by_date'].present?
+  def batch_update
+    params['hold_list'].each do |hold_key|
+      pass_these = { hold_key: hold_key,
+                     pickup_library: params['pickup_library'],
+                     session_token: current_user.session_token }
```
banukutlu, line 25: can we pass `symphony_client` as well? would it make sense 🤔

app/jobs/change_pickup_library_job.rb
```diff
@@ -0,0 +1,16 @@
+class ChangePickupLibraryJob < ApplicationJob
+  queue_as :default
+
+  def perform(pass_these)
+    symphony_client = SymphonyClient.new
```
banukutlu, line 5: Can we also pass `symphony_client` with `pass_these`? So there won't be need for this line `symphony_client = SymphonyClient.new` here. Would it make sense to pass it? 🤔

app/services/symphony_client.rb
```diff
@@ -220,7 +220,14 @@ def request(path, headers: {}, method: :get, **other)
         .headers(default_headers.merge(headers))
         .request(method, base_url + path, **other)
     end
-
+  # def request(path, headers: {}, method: :get, **other)
+  # logger = Logger.new(STDOUT)
+  # HTTP
+  #     .use(logging: { logger: logger },
```
banukutlu, line 226: curious why you added then commented out logging? did not work as you expect or decided not to use?

app/views/holds/_holds_ready_table.html.erb
```diff
@@ -1,3 +1,4 @@
+
```
banukutlu, line 1:
```suggestion
```

config/routes.rb
```diff
@@ -3,9 +3,16 @@
 Rails.application.routes.draw do
   root to: 'sessions#index'
 
+  require 'sidekiq/web'
+  Sidekiq::Web.set :sessions, false
+  mount Sidekiq::Web => '/sidekiq'
+
   resources :summaries, :fines, :checkouts, only: [:index]
   resources :renewals, only: [:create]
   resources :holds, only: [:index, :new, :create, :update, :destroy]
```
banukutlu, line 12:
```suggestion
  resources :holds, only: [:index, :new, :create, :destroy]
```

cdmo:
Hey @banukutlu this PR is a mess, wasn't meant to be reviewed beyond like a "what do you think of this approach?" I plan on working on this more to clean it up and then will ask for a proper review. Thanks for taking a look. Let me know if you have thoughts about my longer plan I mentioned above?

banukutlu:
> Just some thinking outloud:
> 
> * We'll definitely be doing async

yes ✔️  

> * We'll probably use sidekiq (with redis) because of long-running jobs and wanting to avoid timeouts (even in async js calls they'd 500 for users with lots of stuff, like a 500 item checkouts request takes 45 seconds) - this is to say i don't think we can do this without sidekiq. Sidekiq also gives the added [benefit of exponential back off retries](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry). The cost of sidekiq is the overhead of running another tool and the complexity of configuring and deploying it.

💯 agree, `long-running jobs` issue itself could be our reason for going with sidekiq and also error handling and retries are definitely added benefits which would allow us to get rid of the workarounds we added for WS calls that have records in use and item nil issues so even get rid of item and hold lookup calls.

> * We probably don't need to "asyncify" all the SymphonyClient calls. Like logging in and your new stale check. The profile information calls are also very efficient. I think we can get away with not asyncifying fines, but we probably should just in case there are some very heavily fined patrons out there.
> * We'll have to asyncify the display of holds, checkouts/renewals.

I agree that we do not need to asyncify the login and ping checks, also bib, item and hold info lookups (we might even not need hold and item lookup anyway) but most of the rest seems to be done, we can go over calls and decide.

banukutlu:
> Hey @banukutlu this PR is a mess, wasn't meant to be reviewed beyond like a "what do you think of this approach?" I plan on working on this more to clean it up and then will ask for a proper review. Thanks for taking a look. Let me know if you have thoughts about my longer plan I mentioned above?

no worries, I looked at the general approach with knowing that you haven't done cleaning up yet as this is POC, yeah I added some suggestions to delete some lines but it is not with the intention of requesting changes, it was just easy to click and suggest :) 

I added some thoughts/replies to your long comment, but overall I think this is the way we should go and I like your general here. This will help us so much to give results back w/o making the user wait for the whole process to finish which is important for long-running jobs like renewals/cancelling/etc. for many items.

app/jobs/change_pickup_library_job.rb
cdmo: Yup!

app/services/symphony_client.rb
cdmo: Because I don't always want it on, it's just here so I can swap it with the other one when i want more data

app/jobs/change_pickup_library_job.rb
cdmo: Tried to do this and got a "Serialization" error. Apparently you can't pass certain objects to an activejob?

spec/controllers/holds_controller_spec.rb
```diff
@@ -117,7 +96,7 @@
         end
       end
 
-      context 'when not_needed_after param is sent and the webservice is not responding with 200' do
+      xcontext 'when not_needed_after param is sent and the webservice is not responding with 200' do
```
cdmo, line 99: Skipping these for now as I plan to rewrite in another PR

app/controllers/holds_controller.rb
```diff
@@ -18,14 +18,15 @@ def index
...
+    params['hold_list'].each do |hold_key|
+      ws_args = { hold_key: hold_key,
+                  pickup_library: params[:pickup_library],
+                  pickup_by_date: params[:pickup_by_date],
+                  session_token: current_user.session_token }
+      ChangePickupLibraryJob.perform_later(ws_args) if params[:pickup_library].present?
+      handle_not_needed_after_request(ws_args) if params[:pickup_by_date].present?
     end
-
-    redirect_to holds_path
```
cdmo, line 28: Rails responds with a "204" no changes. This OK, I think. There are some other options like `render head: ok` or something like that

app/controllers/holds_controller.rb
```diff
@@ -149,7 +135,7 @@ def handle_not_needed_after_request
     end
 
     def process_flash(type, translation)
-      flash[type] = "#{flash[type]} #{t "myaccount.hold.#{translation}", bib_summary: @hold_to_act_on.bib_summary}"
```
cdmo, line 152: this was removed just to get rails to stop complaining, it'll be fixed in a PR soon

config/webpack/environment.js
```diff
@@ -1,5 +1,7 @@
 const { environment } = require('@rails/webpacker')
+const erb =  require('./loaders/erb')
```
cdmo, line 2: I didn't end up using this. Guess I'll remove it.

app/javascript/submission_handling/polling.js
```diff
@@ -0,0 +1,20 @@
...
+        return result;
+    } catch {
+        reportError();
+    }
+}
+
+export const reportError = function() {
+    // This will be caught as an error by ruby's http gem and reported in logs as well.
+    // Thinking it'd be good to replace this with real content. Probably not an alert.
+    alert("There was a problem contacting the Libraries' lending services. Please call 555-555-5555 for help or try again..");
```
cdmo, line 19: marked for further review. We might need to do something fancier? Like a "toast"? Or a modal? Could set a Rails Flash, but not sure how that would work being called from JS.

app/views/holds/_holds_not_ready_table.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with url: '/holds/multiple', method: :delete do %>
+<%= form_with url: '/batch_update', method: :delete, id: 'pending-holds' do %>
```
cdmo, line 1: changed the path/method name just to make it clearer in my head

app/views/layouts/application.html.erb
```diff
@@ -57,7 +57,9 @@
       <div class="row">
         <div class="col-md-12">
           <%= render partial: '/shared/flash_msg', layout: nil %>
-          <%= yield %>
+          <div id="page-content">
```
cdmo, line 60: a unique class I added at some point for some reason... 🤔

config/routes.rb
```diff
@@ -3,9 +3,16 @@
 Rails.application.routes.draw do
   root to: 'sessions#index'
 
+  require 'sidekiq/web'
+  Sidekiq::Web.set :sessions, false
+  mount Sidekiq::Web => '/sidekiq'
```
cdmo, line 8: Not sure if this is okay to have in production.

cdmo: It's a sidekiq dashboard btw if you haven't checked it yet

cdmo:
Added #179 and #180 as left to do

cdmo:
CodeClimate is not happy because I didn't fix the pickup date change too, will do in another PR, this PR is kinda too big

config/routes.rb
banukutlu: yup 👍

config/webpack/environment.js
banukutlu: 👍  I guess, this to be removed in another PR?

cdmo: I'll remove it from this one

app/views/holds/_holds_not_ready_table.html.erb
banukutlu: what do you say for sth like `/holds/batch_update`?

cdmo: yeah that's better, I'll change it

cdmo: gotta update the route too and stuff, more to it

app/javascript/submission_handling/change_pickup_library.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Attributes
+const pendingHoldsForm = document.querySelector('form#pending-holds');
+const pendingHoldsFormCheckboxes = document.querySelectorAll('form#pending-holds .checkbox');
+const spinner = `<div class="spinner-border" role="status">
+                   <span class="sr-only">Loading...</span>
+                 </div>`;
+const pickupChangeSelect = document.querySelector('[data="pickup-location"]');
+
+const getJobInfo = async function (holdId) {
```
banukutlu, line 14: @cdmo do you think we can take methods like `getJobInfo` and `deleteData` out and move to a file to be used by other asyncing work as well, like polling.js idea? These seems to gonna be similar among all?
If this is sth that makes sense, we do need to do this with this PR since this is already a big PR and this is very close to be merged so we can do that in another PR.

cdmo: I think that's a great suggestion. I can do it in this PR, np.

cdmo: ok that I moved it into polling.js you think? they seem tied to me. The way I did it is that the function is just expecting you to give it a callback from whatever is calling it, and that callback will probably do some DOM manipulation (like updating the pickup library `td` cell)

cdmo:
I added the feature spec since it was done. These informed the work:

* https://github.com/mperham/sidekiq/wiki/Testing#capybara-feature-tests-with-backend-running-in-parallel
* https://docs.travis-ci.com/user/database-setup/#redis
* https://github.com/wardencommunity/warden/blob/5b3cbd5bef67cbe399bb7007537bc4841bbee772/lib/warden/test/helpers.rb#L14

Shockingly the feature spec worked on the first try with Travis

app/javascript/submission_handling/change_pickup_library.js
banukutlu: yes, polling.js works ✔️

app/javascript/submission_handling/polling.js
```diff
@@ -0,0 +1,50 @@
...
+    } catch (error) {
+        reportError(error);
+    }
+
+    return result;
+};
+/* eslint-enable max-statements */
+
+const getJobInfo = async function (holdId) {
+    let response = await fetch(`/redis_jobs/${holdId}`);
```
banukutlu, line 37:
```suggestion
const getJobInfo = async function (jobId) {
    let response = await fetch(`/redis_jobs/${jobId}`);
```

app/javascript/submission_handling/polling.js
```diff
@@ -0,0 +1,50 @@
...
+
+    return response.json();
+};
+
+const deleteData = function (item) {
+    fetch(`/redis_jobs/${item}`, { "method": "delete" });
```
banukutlu, line 43:
```suggestion
const deleteData = function (jobId) {
    fetch(`/redis_jobs/${jobId}`, { "method": "delete" });
```

just for consistency with the getJobInfo method?

app/javascript/submission_handling/polling.js
```diff
@@ -0,0 +1,50 @@
...
+    let response = await fetch(`/redis_jobs/${jobId}`);
+
+    return response.json();
+};
+
+const deleteData = function (jobId) {
+    fetch(`/redis_jobs/${jobId}`, { "method": "delete" });
+};
+
+export const renderData = async function (target, checkResults, resultCallback) {
```
banukutlu, line 46: 👌

app/javascript/submission_handling/polling.js
```diff
@@ -0,0 +1,50 @@
...
+    const maxWaitTime = 20000;
+    const pollInterval = 1000;
+    const endTime = Number(new Date()) + maxWaitTime;
+    let result = '';
+
+    try {
+        result = await fn(arg);
+        /* eslint-disable no-await-in-loop */
+        while (checkFn(result) && Number(new Date()) < endTime) {
+            await sleep(pollInterval);
```
banukutlu, line 24: 👌

spec/features/holds_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    login_as(username: 'PATRON1', patron_key: mock_user)
+  end
+
+  after do
+    Redis.new.del('item_type_map')
+  end
+
+  it 'lets the user change the pickup library of a hold', js: true do
+    visit holds_path
+    login_as(username: 'PATRON1', patron_key: mock_user)
```
banukutlu, line 18: interesting that it requires another login_as, is it like per request or sth?

cdmo: yup. It does. the login_as will work for *the next page being viewed (like http request)* which i learned the hard way through very cryptic error messages. The line I linked to from the warden repo is what clued me in. https://github.com/wardencommunity/warden/blob/5b3cbd5bef67cbe399bb7007537bc4841bbee772/lib/warden/test/helpers.rb#L14

PR #168, opened 2020-03-31
rails upgrade so to upgrade actionview

banukutlu:
To address [this](https://github.com/psu-libraries/myaccount/network/alert/Gemfile.lock/actionview/open)


cdmo:
don't we have to change the naming for view components then too?

cdmo:
oh nevermind, sorry

PR #170, opened 2020-04-06
refactoring stale session check

banukutlu:


banukutlu:
> Talked IRL about authenticate ws call

I changed the authenticate request to patron check with including only a key field, and it seems to be even slightly faster than authenticate one :)

app/services/symphony_client.rb
```diff
@@ -32,12 +32,9 @@ def login(user_id, password)
   end
 
   def authenticate(user)
-    request('/user/patron/authenticate',
-            headers: { 'x-sirs-sessionToken': user.session_token }, method: :post,
-            json: {
-              barcode: user.username,
-              password: user.patron_key
-            })
+    authenticated_request("/user/patron/key/#{user.patron_key}",
```
cdmo, line 35: Looks great, this is a nice solution.

I would only recommend changing the method name from `def authenticate` to `def ping` or something like that. `def session_alive?` maybe? That would mean converting return values to true/false (200 = true, anything else is false). Just some ideas.

cdmo: I think adding a one line comment to the method might be good too.

banukutlu: > Looks great, this is a nice solution.
> 
> I would only recommend changing the method name from `def authenticate` to `def ping` or something like that. `def session_alive?` maybe? That would mean converting return values to true/false (200 = true, anything else is false). Just some ideas.

I like `ping` 👍 will make the change and the return values change

app/services/symphony_client.rb
```diff
@@ -31,10 +31,13 @@ def login(user_id, password)
...
-    authenticated_request("/user/patron/key/#{user.patron_key}",
-                          headers: { 'x-sirs-sessionToken': user.session_token },
-                          params: { includeFields: 'key' })
+  # This method is for validating user session_token
+  def ping?(user)
+    response = authenticated_request("/user/patron/key/#{user.patron_key}",
+                                     headers: { 'x-sirs-sessionToken': user.session_token },
+                                     params: { includeFields: 'key' })
+
+    response.status == 200
```
cdmo, line 40: I'm glad you moved the `response.state == 200` part up here, makes the calling of this method simpler (don't have to do an additional step everytime it's called).

spec/services/symphony_client_spec.rb
```diff
@@ -19,11 +19,33 @@
         .to_return(body: { patronKey: Settings.symws.patron_key }.to_json)
     end
 
-    it 'authenticates the user against symphony' do
+    it 'logs the user in to symphony' do
```
cdmo, line 22: Oh this is a good change, match the language of the method

spec/services/symphony_client_spec.rb
```diff
@@ -71,6 +93,84 @@
     end
   end
 
+  describe '#change_pickup_library' do
```
cdmo, line 96: Kinda curious as to why you added all of these SC specs for old methods. I didn't realize we were missing specs here. Thanks for adding them, I would definitely have done them though if you asked :). Looks like we are still missing one, `get_item_type_map`. I can take care of that one.

Issue #179, opened 2020-04-17
Web service error logging and in-page error notifications to users


cdmo:
Create contents of the polling.js error function

- [x] Client side error notifications
- [x] ~Client side error logging~ see #207 
- [x] Server side error logging

from #166

cdmo:
This came up at our checkin today, we and @ruthtillman all seemed to agree that using a Bootstrap Toast is probably a good idea to try as long as it can be canceled and is accessible

cdmo:
I rigged up an example of how this will work. I think this is working for the most part, I do like it. It's less obtrusive than a modal window or an alert which is nice. Here's a demo, @ruthtillman what do you think? (apologies for poor image quality):

![toast](https://user-images.githubusercontent.com/523381/80033364-9366e900-84ba-11ea-930f-503972f53360.gif)

ruthtillman:
Yes, I like the obvious redness when there's a problem and then the Redness of error.

If they're successful, will it still load the green box at the top or anything like that?

ruthtillman:
I think if it's successful, then anything that needs to be closed like a toast is probably overkill, so loading in the older way would be better?

cdmo:
toasts can be set to auto close after a duration. I was thinking that 10 seconds would be a reasonable amount of time for any toast to be closed (success or fail). For someone with a lot of items the toasts could pile up on a big requests, but not sure what else can be done about that. Toasts can also be positioned anywhere on the screen fwiw.

The older way could work okay, but, if a user is scrolled down at the bottom of the page adding a message at the top might be missed.

It's kinda tricky.

cdmo:
Recap what we discussed at today's check in meeting:

- [ ] We'll only use toasts for errors
- [ ] Success will be shown as green checks  ✅  next to the update item after the spinner is gone and the new value is present.
- [ ] We'll need to make sure to translate errors as needed (as in error codes from Sirsi translated to human-understandable messages)
- [ ] Monitoring will be critical

banukutlu:
just attaching a screenshot, it looks nice overall

<img width="1051" alt="Screen Shot 2020-07-02 at 7 23 33 PM" src="https://user-images.githubusercontent.com/6477867/86416983-aad00880-bc99-11ea-8c9d-7a033ace092a.png">

banukutlu:
[...]

PR #182, opened 2020-04-22
Toast (and general error handling of api calls)

cdmo:
#179
#198


cdmo:
Just a note that we should change all error messages in logs to be raw message form sirsi ws.

app/javascript/submission_handling/polling.js
```diff
@@ -46,7 +51,9 @@ const pollFetch = function(arg, otherRule = null) {
         }).
         catch((error) => {
             // This would be a network error
-            reportError(error);
+            // @todo: create a logging service
```
cdmo, line 54: This really needs to get done I think. Yet another reason to try [honeybadger](https://docs.honeybadger.io/lib/javascript/index.html). Or we can roll our own.

app/services/sirsi_response/error.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def wrangle_symphony(bib_type, symphony_client, session_token, key)
+      if bib_type == :checkout
+        return SymphonyClientParser::parsed_response(symphony_client, :get_item_info, nil, session_token, key)
+      end
+
+      SymphonyClientParser::parsed_response(symphony_client, :get_hold_info, key, session_token)
+    end
+end
```
cdmo, line 44: This service does not have a specific spec but is tested via integration (job specs). Not sure if we should add a specific `error_spec.rb` or not. Would like your feedback. We probably need a spec for this right?

cdmo:
This is ready to go, lots of little changes combined with some big ones, but the idea is that we now have a common way of handling all client side errors (server logging and client side notifications via toasts). Still would like to add client side error logging, but that's a discussion I think.

cdmo:
CodeClimate marked this as failing because of some issues it found. You can either choose to mark them "ok" or we can try and fix them as part of the PR review. Your call.

app/javascript/submission_handling/polling.js
```diff
@@ -59,11 +66,22 @@ const deleteData = function (jobId) {
...
-        // This would be a network error
-        reportError(error);
+        let genericError = { "new_value_formatted": "Error",
+                             "id": target };
+        resultCallback(genericError);
+        reportError('There was a network error, please try again later or call your campus library.', target);
+        // The max wait time was reached. Web Service is probably down.
+        // @todo: create a logging service
+        // eslint-disable-next-line no-console
+        console.error(error);
```
banukutlu, line 84:
```suggestion
```

app/jobs/cancel_hold_job.rb
```diff
@@ -11,17 +11,24 @@ def perform(hold_key:, session_token:)
     case response.status
     when 200
       Redis.current.set("cancel_hold_#{hold_key}", {
-        hold_id: hold_key,
-        result: :success
+        id: hold_key,
```
banukutlu, line 14: I liked using `id` for all 👍

app/jobs/cancel_hold_job.rb
```diff
@@ -11,17 +11,24 @@ def perform(hold_key:, session_token:)
...
+        result: :success,
+        response: '<p class=\'text-danger\'>Cancelled</p>'
```
banukutlu, line 16: 👍

app/jobs/cancel_hold_job.rb
```diff
@@ -11,17 +11,24 @@ def perform(hold_key:, session_token:)
...
       }.to_json)
     else
-      error_message_raw = JSON.parse response.body
-      error_message = error_message_raw&.dig('messageList')&.first&.dig('message') || 'Something went wrong'
-      Sidekiq.logger.error("cancel_hold_#{hold_key}: #{error_message}")
+      processed_error = SirsiResponse::Error.new(error_message_raw: JSON.parse(response.body),
```
banukutlu, line 19: 👌  nice

app/services/sirsi_response/error.rb
```diff
@@ -0,0 +1,44 @@
...
+                                       layout: false,
+                                       locals: { id: key,
+                                                 title: bib_record.title,
+                                                 display_error: display_error }
+  end
+
+  private
+
+    def make_human_friendly_response
+      SIRSI_RESPONSE_TRANSLATIONS[@log&.dig('messageList')&.first&.dig('code')] ||
```
banukutlu, line 32: these translations are only for renewals, is it like that here? @cdmo

app/jobs/change_pickup_library_job.rb
```diff
@@ -16,19 +16,30 @@ def perform(hold_key:, session_token:, pickup_library:)
...
+      Sidekiq.logger.error("pickup_library_#{hold_key}: #{processed_error.log}")
+
       Redis.current.set("pickup_library_#{hold_key}", {
-        hold_id: hold_key,
+        id: hold_key,
         result: :failure,
-        response: error_message
+        response: {
+          new_value: 'Error',
+          new_value_id: 'Error'
```
banukutlu, line 40: `new_value_id: 'Error'` is this supposed to be `Error`?

app/jobs/change_pickup_library_job.rb
```diff
@@ -16,19 +16,30 @@ def perform(hold_key:, session_token:, pickup_library:)
...
+        response: {
+          new_value: human_readable_location,
+          new_value_id: pickup_library
+        }
       }.to_json)
     else
-      error_message_raw = JSON.parse response.body
-      error_message = error_message_raw&.dig('messageList')&.first&.dig('message') || 'Something went wrong'
-      Sidekiq.logger.error("pickup_library_#{hold_key}: #{error_message}")
+      processed_error = SirsiResponse::Error.new(error_message_raw: JSON.parse(response.body),
```
banukutlu, line 27: Wondering if we can DRY these jobs by adding a concern or some other way?

app/jobs/renew_checkout_job.rb
```diff
@@ -31,29 +22,34 @@ def perform(resource:, item_key:, session_token:)
...
 
       Redis.current.set("renewal_#{item_key}", {
-        item_key: item_key,
+        id: item_key,
         result: :failure,
-        error_message: renewal_error_message(response)
+        response: {
+          renewal_count: 'Error',
+          due_date: 'Error',
+          status: 'Error'
```
banukutlu, line 49: What about rather keeping the previous values? It just looks there are too many `Error`s on the page when there many renewals attempted :) Just a thought..

app/services/sirsi_response/error.rb
cdmo: No, it's for any errors, but it doesn't hurt to check. I imagine we may find more in the future in holds requests? Or if other errors showup. If not we can add a check for bib type

app/jobs/change_pickup_library_job.rb
cdmo: Nope. This is the value that gets processed by javascript as the new value in the table of holds. It'll say "Error" instead of whatever the new value should have been. Though I can't generate hold errors anymore infuriatingly.

app/javascript/submission_handling/polling.js
cdmo: I'd rather keep this until we replace it with a logging service. console.logs (errors, etc) are bad if kept because they are used for debugging, but this is for error reporting. I am reading into your suggestion though, maybe incorrectly. Also, there is a second console.error - did you see that one too?

app/jobs/renew_checkout_job.rb
cdmo: Yeah, I thought about that too. Probably better? Maybe like providing the current values but with a warning sign thingy like a exclamation point in a triangle next to it? Idk. Good feedback. We can ask Ruth too.

app/services/sirsi_response/error.rb
banukutlu: Afair we were gonna keep the other error messages as the way sirsi has, currently the cat only have these for renewals, we can check with Ruth, if so adding a check for bib type makes sense.

app/jobs/change_pickup_library_job.rb
cdmo: Yeah. I guess the frustrating part for me is that it is DRYed out a bit. This was a stack of procedures being executed in order on every job and I replaced it with a `SirsiResponse::Error`. It just has a ton of arguments, which is why I think CC is complaining. There's probably a clever way to send less args. 🤷

app/jobs/renew_checkout_job.rb
banukutlu: I like the idea of current values with a warning sign icon next to it, yes let's check with Ruth on this.

app/jobs/change_pickup_library_job.rb
cdmo: I was kinda thinking about some other abstractions that I ultimately left out. Like a model for the values we set in Redis instead of using a hash in a consistent manner. Thought about a lookup_title kind of utility module that could be mixed in with any job. Like this was the module:

```rb
module SirsiResponse::LookupTitle
  def get_title(symphony_client:, bib_type:, session_token:, identifier:)
    response = if bib_type == :checkout
                 SymphonyClientParser::parsed_response(symphony_client,
                                                       :get_item_info,
                                                       nil, session_token, identifier)
               else
                 SymphonyClientParser::parsed_response(symphony_client,
                                                       :get_hold_info,
                                                       key, session_token)
               end

    response.title
  end
end
```

Maybe that's better I don't know. I think I just went with this Error class instead because it is responsible for formatting and setting errors (which includes getting the title).

app/services/sirsi_response/error.rb
cdmo: Okay, that sounds good. Thanks. Guess we could also abstract this Error class a bit more and then make subclasses for CheckoutError and HoldError.

app/javascript/submission_handling/polling.js
banukutlu: ok cool, I just realized you even added a todo there 👍

app/jobs/change_pickup_library_job.rb
banukutlu: this is also related to this https://github.com/psu-libraries/myaccount/pull/182#discussion_r449353149, right?

app/jobs/renew_checkout_job.rb
banukutlu: fyi asked for Ruth's feedback here https://github.com/psu-libraries/myaccount/issues/179#issuecomment-653579690

app/services/sirsi_response/error.rb
banukutlu: checking with Ruth https://github.com/psu-libraries/myaccount/issues/179#issuecomment-653581309

app/services/sirsi_response/error.rb
banukutlu: I think I am leaning towards adding a specific spec for it 🤔

app/javascript/submission_handling/polling.js
banukutlu: I agree! honeybadger would be neat! We have #207 for this right?

app/jobs/change_pickup_library_job.rb
banukutlu: > Yeah. I guess the frustrating part for me is that it is DRYed out a bit. This was a stack of procedures being executed in order on every job and I replaced it with a `SirsiResponse::Error`. It just has a ton of arguments, which is why I think CC is complaining. There's probably a clever way to send less args. 🤷

yes there are def. great improvements you implemented here 👌

banukutlu: > I was kinda thinking about some other abstractions that I ultimately left out. Like a model for the values we set in Redis instead of using a hash in a consistent manner. Thought about a lookup_title kind of utility module that could be mixed in with any job. Like this was the module:
> 
> Maybe that's better I don't know. I think I just went with this Error class instead because it is responsible for setting errors (which includes getting the title).

Error class makes so much sense to me ✔️  As you mentioned just finding a way to send lesser arg.s might do make CC happy 🤔

app/javascript/submission_handling/polling.js
```diff
@@ -59,11 +66,22 @@ const deleteData = function (jobId) {
...
         resultCallback(result);
         deleteData(target);
     }).
     catch((error) => {
-        // This would be a network error
-        reportError(error);
+        let genericError = { "new_value_formatted": "Error",
+                             "id": target };
+        resultCallback(genericError);
+        reportError('There was a network error, please try again later or call your campus library.', target);
```
banukutlu, line 80:
```suggestion
        reportError('There was a network error, please try again later or call your campus library.');
```

is target needed here?

app/javascript/submission_handling/polling.js
```diff
@@ -59,11 +66,22 @@ const deleteData = function (jobId) {
 
 export const renderData = (target, resultCallback, otherRule = null) => {
     pollFetch(target, otherRule).then((result) => {
+        if (checkError(result)) {
+            reportError(result.display_error, result.id);
```
banukutlu, line 70:
```suggestion
            reportError(result.display_error);
```
is `result.id` needed?

app/jobs/change_pickup_library_job.rb
cdmo: No, that comment is in regards to what gets sent back to the client as `display_error` (what shows up in the toast)

app/services/sirsi_response/error.rb
cdmo: I think it's nice to have the option to override whenever we want, and doing this doesn't cost much and keeps things simpler. 🤷

app/services/sirsi_response/error.rb
cdmo: Yeah. Think you're right.

app/javascript/submission_handling/polling.js
cdmo: Yeah I just created that issue to have a full discussion on this topic yup. Would love for a product to do this job so nicely for us. Probably won't happen though, so we'll likely have to create our own.

app/jobs/change_pickup_library_job.rb
cdmo: Would be great if the symph client instance and the token were globally accessible somehow, or modularized as a mixin to be included whereever needed. Just not sure it would work out.

app/javascript/submission_handling/polling.js
cdmo: Nope! Great catch. Surprise js-lint doesn't have the ability to catch this.

Thanks!

app/javascript/submission_handling/polling.js
```diff
@@ -38,6 +41,8 @@ const pollFetch = function(arg, otherRule = null) {
         getJobInfo(arg).then((data) => {
             if (validateResult(data, otherRule)) {
                 resolve(data);
+            } else if (checkError(data)) {
+                resolve(data);
```
banukutlu, line 45: 👍

app/jobs/change_pickup_library_job.rb
banukutlu: Oops sorry added the wrong the link, meant to link to this convo https://github.com/psu-libraries/myaccount/pull/182#discussion_r449353998

banukutlu: So, basically if we decide to keep/send the current values, these would not have the value `Error`, is what I was trying to ask :)

cdmo: Yeah, this part would need changing/rethinking

cdmo:
I'd like to handle the question of changing the error presented in the tables to it's original value to a separate issue if that's alright with you? We may need to talk it through in a checkin with ruth is what I'm thinking

cdmo:
I don't know why Travis failed. rubocop passes locally for me.

spec/services/sirsi_response/error_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      it 'logs the server response from Sirsi' do
+        expect(error.log).to include 'messageList' => [{ 'code' => 'Not a real code 1234',
+                                                         'message' => 'Some error message' }]
+      end
+    end
+
+    context 'when an error for a hold occurs that is malformed' do
+      let(:error_message) { nil }
+
```
banukutlu, line 42:
```suggestion
      let(:bib_type) { 'hold' }
      let(:error_message) { nil }
```

cdmo: No because this particular context is testing for a malformed error which should be treated the same regardless of `bib_type` (it'll hit that last option in the error message translation method). Or am I wrong? Quite likely! :)

cdmo:
I am retrofitting this to change how the table values are treated as per our discussions. Changing this to WIP

spec/spec_helper.rb
```diff
@@ -99,7 +99,7 @@
   #   Kernel.srand config.seed
 
   # Mock all requests to symphony for feature tests
-  config.before type: :feature do
+  config.before :example, type: Proc.new { :service || :feature } do
```
banukutlu, line 102: `config.before type: :feature do` not working? I did not know this 🤔

spec/services/sirsi_response/error_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let(:mock_client) { SymphonyClient.new } # Creates actual object, stubs API responses with Sinatra
+    let(:key) { '123457' }
+    let(:session_token) { '92929292' }
+    let(:bib_type) { 'checkout' }
+
+    after do
+      Redis.current.flushall
+    end
+
+    context 'when an error for a hold occurs that does not have a translation' do
```
banukutlu, line 29:
```suggestion
    context 'when an error occurs that does not have a translation' do
```
since this spec is any context regardless of bib_type?

spec/spec_helper.rb
cdmo: It is working, I just wanted to add the sinatra fake web service to service specs too so I didn't have to do a whole bunch of redundant mocking

spec/services/sirsi_response/error_spec.rb
cdmo: yes yes

spec/spec_helper.rb
banukutlu: ahhh yes cool! I just did not realize the service part :facepalm:

cdmo:
This is not ready to merge, but I would like you to try out a checkout renew that generates a failure and see how it goes. If it is good, I'll go through and update the way inline errors are displayed for the rest of the async options as well, but I want to make sure it's good so that I don't waste effort.

banukutlu:
> This is not ready to merge, but I would like you to try out a checkout renew that generates a failure and see how it goes. If it is good, I'll go through and update the way inline errors are displayed for the rest of the async options as well, but I want to make sure it's good so that I don't waste effort.

sure, will do

app/javascript/submission_handling/renew_checkout.js
```diff
@@ -1,12 +1,20 @@
-import { allChecked, findForm, responseFromRails, spinner } from './shared'
+import { allChecked, findForm, responseFromRails, toggleSpin } from './shared'
```
banukutlu, line 1: `toggleSpin` 👍  nicer and neat

app/jobs/change_pickup_by_date_job.rb
```diff
@@ -34,10 +34,6 @@ def perform(hold_key:, session_token:, pickup_by_date:)
       Redis.current.set("pickup_by_date_#{hold_key}", {
         id: hold_key,
         result: :failure,
-        response: {
-          new_value: 'Error',
-          new_value_formatted: 'Error'
-        },
```
banukutlu: 👌

app/jobs/renew_checkout_job.rb
```diff
@@ -43,11 +43,6 @@ def perform(resource:, item_key:, session_token:)
       Redis.current.set("renewal_#{item_key}", {
         id: item_key,
         result: :failure,
-        response: {
-          renewal_count: 'Error',
-          due_date: 'Error',
-          status: 'Error'
-        },
```
banukutlu: I really liked that these are not needed anymore 👍

app/views/checkouts/_checkouts_table.html.erb
```diff
@@ -29,9 +29,9 @@
             <td>
               <%= render BibItemComponent, bibitem: checkout %>
             </td>
-            <td class="renewal_count"><%= checkout.renewal_count if checkout.renewal_count.positive? %></td>
-            <td class="due_date"><%= render partial: 'checkouts/due_date', locals: { checkout: checkout } %></td>
-            <td class="status"><%= checkout.status_human %></td>
+            <td class="renewal_count"><span><%= checkout.renewal_count if checkout.renewal_count.positive? %></span></td>
```
banukutlu, line 32: just curious why needed a `span` here, I am sure there is a reason just trying to understand

banukutlu:
> This is not ready to merge, but I would like you to try out a checkout renew that generates a failure and see how it goes. If it is good, I'll go through and update the way inline errors are displayed for the rest of the async options as well, but I want to make sure it's good so that I don't waste effort.

I think this looks very good, I tried a renewal and also did an overview review. I liked your approach and how you refactored the toggle_spin very much 👌   Will you make a change to display less warning triangles on the renewal, sth Ruth brought up I guess, right?

cdmo:
yup, and will update error display for holds too. thanks!!

cdmo:
:hurtrealbad: ready. for. review.

app/views/checkouts/_checkouts_table.html.erb
banukutlu: ok makes sense, thanks :)

spec/features/checkouts_spec.rb
```diff
@@ -25,4 +25,14 @@
         .and have_css '[id="checkout2145643:5:1"] .status', text: ''
     end
   end
+
+  context 'when patron fails to renew a checkout successfully' do
```
banukutlu, line 29: 🎉 nice!

spec/views/checkouts/all.html.erb_spec.rb
```diff
@@ -36,19 +36,19 @@
...
+      expect(rendered).to include '<td class="renewal_count"><span>2</span></td>'
     end
 
     it 'displays checkout\'s estimated overdue amount when item is overdue' do
       expect(rendered).to include '$10.00'
     end
 
     it 'displays checkout\'s due date correctly when recalled' do
-      expect(rendered).to include 'Recalled&lt;br&gt;December 20, 2019&lt;br&gt;November 14, 2019'
+      expect(rendered).to match /(Recalled<br>)\s*(December 20, 2019<br>)\s*November 14, 2019/
```
banukutlu, line 47: ✔️

PR #184, opened 2020-04-26
181 pickup date

cdmo:
#181 

Includes some refactors as well

also I have a feeling i know what was going on with turbolinks too, will test in a separate branch

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -20,12 +20,15 @@ class HoldsController < ApplicationController
   # PUT /holds
   def batch_update
     params['hold_list'].each do |hold_key|
-      ws_args = { hold_key: hold_key,
-                  pickup_library: params[:pickup_library],
-                  pickup_by_date: params[:pickup_by_date],
-                  session_token: current_user.session_token }
-      ChangePickupLibraryJob.perform_later(ws_args) if params[:pickup_library].present?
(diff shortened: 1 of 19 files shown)


cdmo:
Well I've made 2 total feature specs and I'm already in hot water here! Ha, travis doesn't want to use selenium headless (firefox) so I'd have to set that up. Alternatively selenium chrome headless has this fun failure:


```
Failures:

  1) Holds when a patron has some holds not yet ready to pickup (i.e., pending) lets the user change the pickup by date of a hold
     Failure/Error: expect(page).to have_css '#hold3911148 .pickup_by', text: 'January 1, 9999'
       expected to find visible css "#hold3911148 .pickup_by" with text "January 1, 9999" but there were no matches. Also found "September 9, 275760", which matched the selector but not all filters.
     # ./spec/features/holds_spec.rb:32:in `block (3 levels) in <top (required)>'

Finished in 6.03 seconds (files took 2.85 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/features/holds_spec.rb:26 # Holds when a patron has some holds not yet ready to pickup (i.e., pending) lets the user change the pickup by date of a hold
```

cdmo:
Ok, spec problem cleared, actually pretty easy fix. This is ready for review.

I see CC has some ideas about cleanup. I might try to make an issue from it in the tool and see what happens.

PR #185, opened 2020-04-27
Feature/ci

whereismyjetpack:
Adds Initial CI configuration. I figured we'd remove travis configuration as part of another PR given that this is successful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,13 +6,8 @@ jobs:
       - image: circleci/ruby:2.6.5-buster-node-browsers
     environment:
       RAILS_ENV: test
-      NVM_DIR: .nvm
     steps:
     - checkout
-    - run:
-        name: "NVM"
-        command: |
-          git clone https://github.com/nvm-sh/nvm.git .nvm
     - restore_cache:
         keys:
           - bundle-{{ checksum "Gemfile.lock" }}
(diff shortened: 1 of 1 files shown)


cdmo:
Thanks for putting this together for us @whereismyjetpack. @banukutlu and I will take a look asap.

cdmo:
Looks like [we can integrate Heroku with CircleCI to do "CD" (continuous delivery)](https://circleci.com/integrations/heroku/) that is, only deploy to a heroku instance if the tests pass.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
+version: 2.1
+
+jobs:
+  test:
+    docker:
+      - image: circleci/ruby:2.6.5-buster-node-browsers
```
cdmo, line 6: Does this one line take care of all browser testing needs (i.e., running selenium_chrome_headless in a feature spec)?

Also, [this](https://circleci.com/docs/2.0/browser-testing/#using-screenshots-and-artifacts) seems like a great selling point

whereismyjetpack: yes. circle does us a kindness and has images that have browsers pre installed. 

`store_artifacts` is pretty rad. right now it's sending coverage up there, you should be able to see it on the job screen.

cdmo: That's really nice. Wrangling that in Travis is a pain.

Notice btw our suite is no longer 7 seconds, now a full minute plus just because of the feature specs. Still not too shabby though.

`store_artifacts` is interesting

banukutlu:
Thanks a lot @whereismyjetpack!

PR #188, opened 2020-04-28
186 delete holds

cdmo:
#181 should be merged first, then I'll rebase pending that review

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -20,12 +20,15 @@ class HoldsController < ApplicationController
   # PUT /holds
   def batch_update
     params['hold_list'].each do |hold_key|
-      ws_args = { hold_key: hold_key,
-                  pickup_library: params[:pickup_library],
-                  pickup_by_date: params[:pickup_by_date],
-                  session_token: current_user.session_token }
-      ChangePickupLibraryJob.perform_later(ws_args) if params[:pickup_library].present?
-      handle_not_needed_after_request(ws_args) if params[:pickup_by_date].present?
+      ws_args = { hold_key: hold_key, session_token: current_user.session_token }
+
+      if params[:pickup_library].present?
+        ChangePickupLibraryJob.perform_later(**ws_args, pickup_library: params[:pickup_library])
+      end
+
+      if params[:pickup_by_date].present?
+        ChangePickupByDateJob.perform_later(**ws_args, pickup_by_date: params[:pickup_by_date])
+      end
     end
   end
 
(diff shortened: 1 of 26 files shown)


cdmo:
Looks like Circle CI ran nice (and was faster than Travis fwiw), thanks @whereismyjetpack

app/controllers/holds_controller.rb
```diff
@@ -20,13 +20,18 @@ def index
   # PUT /holds
   def batch_update
     params['hold_list'].each do |hold_key|
-      ws_args = { hold_key: hold_key,
-                  pickup_library: params[:pickup_library],
-                  pickup_by_date: params[:pickup_by_date],
-                  session_token: current_user.session_token }
-      ChangePickupLibraryJob.perform_later(ws_args) if params[:pickup_library].present?
-      handle_not_needed_after_request(ws_args) if params[:pickup_by_date].present?
+      ws_args = { hold_key: hold_key, session_token: current_user.session_token }
```
banukutlu, line 23: nice and neat with splat 👌

app/controllers/holds_controller.rb
```diff
@@ -20,13 +20,18 @@ def index
...
+
+      if params[:pickup_library].present?
+        ChangePickupLibraryJob.perform_later(**ws_args, pickup_library: params[:pickup_library])
```
banukutlu, line 26: nice and neat with splat 👌

cdmo: ruby is so nice to write 😌

app/javascript/submission_handling/shared.js
```diff
@@ -0,0 +1,24 @@
+export let allChecked = (form) => form.querySelectorAll('.checkbox:checked');
+export let submitterValue = (event) => {
+    if (event.submitter) {
+        return event.submitter.value;
+    }
+
+    return 'Not defined';
+}
+
+export const pendingHoldsForm = document.querySelector('form#pending-holds');
```
banukutlu, line 10: @cdmo shall we make this more generic so can be used by checkouts/renewals as well, or keep them seperate 🤔

app/jobs/cancel_hold_job.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    case response.status
+    when 200
+      redis_client.set("cancel_hold_#{hold_key}", {
+        hold_id: hold_key,
+        result: :success
+      }.to_json)
+    else
+      error_message_raw = JSON.parse response.body
+      error_message = error_message_raw&.dig('messageList')&.first&.dig('message') || 'Something went wrong'
```
banukutlu, line 20: @cdmo why we need a default here? just curios if the response does not always include a message?

banukutlu: for weird cases like when we get bad responses from Sirsi?

spec/features/holds_spec.rb
```diff
@@ -6,19 +6,47 @@
...
+      page.click_button 'Cancel'
+      expect(page).to have_css '#hold3911148 .hold_status', text: 'Cancelled'
+    end
+
+    it 'lets the user cancel a ready for pickup hold', js: true do
+      visit holds_path
+      login_as username: 'PATRON1', patron_key: mock_user
+      page.check 'hold_list__3906718'
+      page.click_button 'Cancel Selected Holds'
+      # sleep 10
```
banukutlu, line 48:
```suggestion
```

no need for this line?

app/javascript/submission_handling/shared.js
cdmo: `submitterValue` and `allChecked` are meant to be generic... you're referring to `pendingHoldsForm` then? 🤔

could do something like `selectForm` that takes an argument of the form `id`? Kinda what you thinking or?

app/jobs/cancel_hold_job.rb
cdmo: Nice catch, I wasn't sure tbh. It was more of a safety than anything.

We're probably going to have to write a small service class or model or something to handle Sirsi error translations I think

spec/features/holds_spec.rb
cdmo: nope, thanks, was using as a debugging method along with changing the selenium driver to not be headless so i could poke around at the DOM and console

app/javascript/submission_handling/shared.js
banukutlu: yes, exactly for pendingHoldsForm so can use it for checkouts form as well. Right, others are generic.

cdmo: 👌 a commit coming

banukutlu: great thanks!

app/controllers/holds_controller.rb
banukutlu: :100:

Issue #189, opened 2020-04-28
Async the getting of holds, displaying holds table(s)


cdmo:

cdmo:
Here's a quick demo of where it is right now 

https://psu.box.com/s/lbjlh5eo5cgd96ehfvx33zj4utzr5l1u

PR #192, opened 2020-04-29
189 display holds async

cdmo:
#189


spec/controllers/holds_controller_spec.rb
```diff
@@ -49,28 +49,22 @@
...
-      expect(response).to render_template 'index'
-    end
+    describe '#index' do
+      before do
+        allow(ViewHoldsJob).to receive(:perform_later)
+      end
 
-    it 'assigns holds ready for pickup' do
-      get :index
+      it 'sends a job to CancelHoldJob' do
```
banukutlu, line 57:
```suggestion
      it 'sends a job to ViewHoldsJob' do
```

spec/views/holds/index.html.erb_spec.rb
```diff
@@ -7,56 +7,12 @@
...
   end
 
-  context 'when there are holds ready to pickup' do
-    before do
-      assign(:holds_ready, [hold])
-    end
+  it 'has a creates a div with a data attribute containing the patron_key' do
+    render
 
-    it 'displays a hold' do
```
banukutlu, line 17: instead of these feature specs for holds display?

spec/views/holds/index.html.erb_spec.rb
```diff
@@ -7,56 +7,12 @@
...
   before do
     allow(hold).to receive(:item_type_mapping).and_return(ITEM_TYPE_MAPPING)
+    assign 'patron_key', '123'
   end
 
-  context 'when there are holds ready to pickup' do
-    before do
-      assign(:holds_ready, [hold])
-    end
+  it 'has a creates a div with a data attribute containing the patron_key' do
```
banukutlu, line 13:
```suggestion
  it 'creates a div with a data attribute containing the patron_key' do
```

spec/views/holds/index.html.erb_spec.rb
```diff
@@ -7,56 +7,12 @@
 
   before do
     allow(hold).to receive(:item_type_mapping).and_return(ITEM_TYPE_MAPPING)
```
banukutlu, line 9:
```suggestion
```

banukutlu: also `let(:hold) { build(:hold) }` on line 6 can be deleted as well, I think

spec/views/holds/index.html.erb_spec.rb
cdmo: yes because they require javascript

spec/views/holds/index.html.erb_spec.rb
cdmo: I think you are right, will change

spec/views/holds/index.html.erb_spec.rb
banukutlu: yeah sure, I did not see any new added so I was meaning to ask will you add any? Or do you think the ones we already have there is already covering display holds as well?

cdmo: right that's what I was thinking, all of the existing specs would fail if this view holds didn't work, it's getting tested implicitly by actions like 'clicking a checkbox'

spec/views/holds/index.html.erb_spec.rb
cdmo: Oh I didn't fix this yet. Will do in next PR

banukutlu: ahhh yes! I forgot that I was waiting on that :) in next year 👍

PR #194, opened 2020-05-06
Sidekiq on heroku with redis

cdmo:
These changes allow for Heroku to work with Redis for everything we use Redis for: caching, Sidkiq and temporary "database"

Key change was making the "database" connection defined once as a global constant.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+worker: bundle exec sidekiq -t 25 -c 2
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bin/rails server -p ${PORT:-5000} -e $RAILS_ENV
-worker: bundle exec sidekiq -t 25 -c 2
+worker: bundle exec sidekiq -t 25
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bin/rails server -p ${PORT:-5000} -e $RAILS_ENV
-worker: bundle exec sidekiq -t 25 -c 20
+worker: bundle exec sidekiq -t 25
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bin/rails server -p ${PORT:-5000} -e $RAILS_ENV
-worker: bundle exec sidekiq -t 25
+worker: bundle exec sidekiq -t 25 -c 20
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
+web: bin/rails server -p ${PORT:-5000} -e $RAILS_ENV
 worker: bundle exec sidekiq -t 25 -c 2
--- a/app/jobs/change_pickup_library_job.rb
+++ b/app/jobs/change_pickup_library_job.rb
@@ -5,7 +5,7 @@ class ChangePickupLibraryJob < ApplicationJob
 
   def perform(hold_key:, session_token:, pickup_library:)
     symphony_client = SymphonyClient.new
-    redis_client = Redis.new
+
     response = symphony_client.change_pickup_library(
       hold_key: hold_key,
       pickup_library: pickup_library,
@@ -15,7 +15,7 @@ class ChangePickupLibraryJob < ApplicationJob
     case response.status
     when 200
       human_readable_location = Hold::PICKUP_LOCATION_ACTUAL[pickup_library.to_sym]
-      redis_client.set("pickup_library_#{hold_key}", {
+      $REDIS_CLIENT.set("pickup_library_#{hold_key}", {
         hold_id: hold_key,
         result: :success,
         new_value: human_readable_location,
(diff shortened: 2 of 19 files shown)


config/initializers/redis.rb
```diff
@@ -0,0 +1 @@
+$REDIS_CLIENT = Redis.new url: Settings.redis.database.uri
```
cdmo, line 1: This is the most important change: 1 global constant defined for 1 connection to Redis (this is used in the "database" redis, i.e., where we store temporary values returned from SC to be updated in the browser and then deleted. This is the one thing that was making heroku fail I think.

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,7 @@
+Sidekiq.configure_server do |config|
+  config.redis = { url: Settings.redis.sidekiq.uri, size: 15 }
```
cdmo, line 2: see http://manuelvanrijn.nl/sidekiq-heroku-redis-calc/

config/puma.rb
```diff
@@ -27,14 +27,14 @@
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
```
cdmo, line 29: Updates taken because of suggestions at https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

app/models/concerns/bib_record.rb
```diff
@@ -62,12 +62,10 @@ def call
...
     # Use Redis to store this globally once an hour
     def item_type_mapping
-      redis = Redis.new
-      return JSON.parse(redis.get('item_type_map')) unless redis.get('item_type_map').nil?
-
-      client = SymphonyClient.new
-      item_type_map = client.get_item_type_map.to_json
-      redis.setex('item_type_map', 3600, item_type_map)
-      JSON.parse(item_type_map)
+      Rails.cache.fetch(['item_type_map', __method__], expires_in: 60.minutes) do
```
cdmo, line 65: Please note I changed this to a cache option (which is still Redis but it is more semantically correct and would still work if we swapped the cache backend)

Procfile
```diff
@@ -0,0 +1,2 @@
+web: bin/rails server -p ${PORT:-5000} -e $RAILS_ENV
```
cdmo, line 1: see https://devcenter.heroku.com/articles/procfile

config/settings.yml
```diff
@@ -4,4 +4,9 @@ symws:
   headers: {}
 home_page_flash_message_html:
 redis:
-  url: <%= ENV.fetch 'REDIS_URL', 'redis://127.0.0.1:6379/1' %>
\ No newline at end of file
+  sidekiq:
```
cdmo, line 7: separate concerns, separate instances. that seems to be the going advice on Redis

https://github.com/mperham/sidekiq/wiki/Using-Redis#multiple-redis-instances

cdmo:
Hey @whereismyjetpack, I know you haven't spent a lot of time in this repo, and there's a bunch of changes in this PR, but I'm hoping you can take a look at some of the system kinds of changes especially. What I mean by system is changes to puma.rb, settings for redis and sidekiq, the procfile, etc. Happy to hear any thoughts/feedback on any of it you might have. Thanks

cdmo:
replaced the global constant with `Redis.current` 

See https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L7

app/models/concerns/bib_record.rb
banukutlu: makes sense!

banukutlu:
> replaced the global constant with `Redis.current`
> 
> See https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L7

nice 👌

app/jobs/change_pickup_library_job.rb
```diff
@@ -5,7 +5,7 @@ class ChangePickupLibraryJob < ApplicationJob
 
   def perform(hold_key:, session_token:, pickup_library:)
     symphony_client = SymphonyClient.new
-    redis_client = Redis.new
+
```
banukutlu, line 8: note to self: remember to make this change for renewals after merging this PR  :)

config/settings.yml
banukutlu: 👍

config/initializers/sidekiq.rb
banukutlu: ah that link is great!

cdmo: yes, I was quite happy to find it :)

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sidekiq.configure_server do |config|
+  config.redis = { url: Settings.redis.sidekiq.uri, size: 15 }
+end
+
+Sidekiq.configure_client do |config|
+  config.redis = { url: Settings.redis.sidekiq.uri, size: 2 }
```
banukutlu, line 8: using the calculator, this seems to be set to 5 but per our conversation only 2 for client might be enough.

PR #195, opened 2020-05-11
Asyncinfy: Renewals

banukutlu:
Issue #177 and Issue #196 

Not included: Handling results and error messages for failed renewals, will be done in another PR, after #182 work completed.


README.md
```diff
@@ -12,7 +12,7 @@
 | Software |  Version |
 |----------|------|
 | `ruby`    |  2.6.5 <br> (_ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]_) |
-| `rails`   |  6.0.1 |
+| `rails`   |  6.0.2 |
```
cdmo, line 15: Thanks for updating this

app/helpers/checkouts_helper.rb
```diff
@@ -2,30 +2,7 @@
...
-  def format_due_date(date)
-    return l(date, format: :long) unless l(date, format: :time_only) == '11:59pm'
-
-    l(date, format: :short)
-  end
-
   def render_renew_button
-    submit_tag 'Renew', class: 'btn btn-primary btn-renewable-submit'
+    submit_tag 'Renew', class: 'btn btn-primary', data: { disable_with: 'Please wait...' }
   end
```
cdmo, line 7: What do you think of removing this helper method too to finish off the complete removal? Looks like it's only used once and doesn't do much special process, just a Rails tag method call.... 🧨

banukutlu: I thought about that, too and could not decide coz the same renew button is used both top and bottom of the table so twice 🤔

cdmo: Yeah I am. The reason is that I think we should avoid using helpers at all if it can be avoided, and the use case for when it is used should include some more complicated logic or processing. I think I'd rather skip the helper and just have two identical submit tags. What do you think?

banukutlu: got the helper method removed 👍

app/controllers/checkouts_controller.rb
```diff
@@ -10,6 +14,21 @@ def index
     @checkouts = checkouts
   end
 
+  # Handles form submission for renewals in Symphony
+  #
+  # PATCH /renewals
+  # PUT /renewals
```
cdmo, line 20:
```suggestion
```

app/jobs/renewal_job.rb
```diff
@@ -0,0 +1,59 @@
...
+    "hatErrorResponse.7703": 'Renewal limit reached, cannot be renewed.',
+    "hatErrorResponse.105": 'Item has been recalled, cannot be renewed.',
+    "hatErrorResponse.252": 'Item has holds, cannot be renewed.',
+    "hatErrorResponse.46": 'Item on reserve, cannot be renewed.',
+    "unhandledException": ''
+  }.with_indifferent_access
+
+  def perform(resource:, item_key:, session_token:)
+    symphony_client = SymphonyClient.new
+    redis_client = Redis.new
```
cdmo, line 17: This Job can change to mimic https://github.com/psu-libraries/myaccount/pull/194/files#diff-949384a9b4789b027b8340913bfe9858 in terms of the way the connection to Redis is handled.

app/jobs/renewal_job.rb
```diff
@@ -0,0 +1,59 @@
...
+  private
+
+    def error_code(response)
+      return if response.status == 200
+
+      JSON.parse(response.body)&.dig('messageList')&.first&.dig('code')
+    end
+
+    def renewal_error_message(response)
+      return if response.status == 200
```
cdmo, line 54:
```suggestion
      return if response.status == 200
```

This seems redundant to me 🤔

cdmo: And if you remove this, might be a good idea to just send `response.message` instead of the whole response object?

app/jobs/renewal_job.rb
```diff
@@ -0,0 +1,59 @@
...
+        result: :failure,
+        error_message: renewal_error_message(response)
+      }.to_json)
+    end
+  end
+
+  private
+
+    def error_code(response)
+      return if response.status == 200
```
cdmo, line 48:
```suggestion
```

app/jobs/renewal_job.rb
```diff
@@ -0,0 +1,59 @@
...
+class RenewalJob < ApplicationJob
+  queue_as :default
+
+  RENEWAL_CUSTOM_MESSAGELIST = {
+    "hatErrorResponse.141": 'Renewal limit reached, cannot be renewed.',
+    "hatErrorResponse.7703": 'Renewal limit reached, cannot be renewed.',
+    "hatErrorResponse.105": 'Item has been recalled, cannot be renewed.',
+    "hatErrorResponse.252": 'Item has holds, cannot be renewed.',
+    "hatErrorResponse.46": 'Item on reserve, cannot be renewed.',
+    "unhandledException": ''
```
cdmo, line 12: For this final case, when does it show up again? I am seeing it when trying to renew items that are overdue. Are there other cases?

Full json:

```
=> {"messageList"=>[{"code"=>"unhandledException", "message"=>"Unhandled exception: class java.lang.NullPointerException"}]}
```

app/jobs/renewal_job.rb
```diff
@@ -0,0 +1,59 @@
...
+        renewal_count: checkout.renewal_count,
+        due_date: checkout.due_date_human,
+        status: checkout.status_human
+      }.to_json)
+    else
+      Sidekiq.logger.error("renewal_#{item_key}: #{renewal_error_message(response)}")
+      redis_client.set("renewal_#{item_key}", {
+        item_key: item_key,
+        result: :failure,
+        error_message: renewal_error_message(response)
```
cdmo, line 40: When the `unhandledException` comes through, then nothing comes through in the logs other than `2020-05-18T22:35:38.273Z pid=60253 tid=ouvxdiup1 class=RenewalJob jid=a4da5f9621eef9a10af1ab5a ERROR: renewal_2089534:3:1:`. I think we need a translation for that error code or we need to dump the response from sirsi here into the logs.

app/models/checkout.rb
```diff
@@ -18,6 +18,25 @@ def status
...
   end
 
+  def status_human
+    return unless overdue?
+
+    claims_returned? ? 'Claims Returned' : 'Overdue'
+  end
+
+  def due_date_human
+    due_dates = []
```
cdmo, line 28: Shouldn't this be singular?

app/models/checkout.rb
```diff
@@ -18,6 +18,25 @@ def status
...
+
+    if recalled?
+      due_dates << 'Recalled'
+      due_dates << format_due_date(recall_due_date)
+    end
+
+    due_dates << format_due_date(due_date)
+
+    due_dates.join('<br>')
```
cdmo, line 37:
```suggestion
  def due_date_human
    format_due_date due_date
  end
  
  def recalled_status
    'Recalled' if recalled?
  end
```

I think we should try to keep any markup out of our models in general, and I'm sort of additionally thinking we should just have this method return the human readable due date (the formatted due date). Then in the renewal job can call recalled_status and due_date_human separately.

app/services/symphony_client.rb
```diff
@@ -12,15 +12,6 @@ class SymphonyClient
...
 
-  RENEWAL_CUSTOM_MESSAGELIST = {
-    "hatErrorResponse.141": 'Renewal limit reached, cannot be renewed.',
-    "hatErrorResponse.7703": 'Renewal limit reached, cannot be renewed.',
-    "hatErrorResponse.105": 'Item has been recalled, cannot be renewed.',
-    "hatErrorResponse.252": 'Item has holds, cannot be renewed.',
-    "hatErrorResponse.46": 'Item on reserve, cannot be renewed.',
-    "unhandledException": ''
-  }.with_indifferent_access
-
```
cdmo, line 23: Like that you moved these to the async job 👍

app/services/symphony_client.rb
```diff
@@ -79,20 +70,17 @@ def not_needed_after(hold_key:, fill_by_date:, session_token:)
...
+    authenticated_request('/circulation/circRecord/renew',
+                          headers: { 'x-sirs-sessionToken': session_token },
+                          method: :post, json: {
+                            item: {
+                              resource: resource,
+                              key: item_key
+                            }
+                          }, params: {
+                            includeFields: 'circRecord{*}'
+                          })
```
cdmo, line 83: This is great too, that it's singular it can be simpler and error handling can be renewal job's concern. I like making symphony client "dumber"

app/javascript/submission_handling/renewal.js
```diff
@@ -0,0 +1,35 @@
+import { allChecked, findForm, responseFromRails, spinner } from './shared'
```
cdmo, line 1: So I see you named this `renewal.js` instead of `checkout.js` - I think I expected everything to be checkouts since you got rid of the renewal controller 🤔

app/javascript/view_renewals.js
```diff
@@ -0,0 +1,7 @@
+import renewal from "./submission_handling/renewal";
+
+const renewals = () => {
+    renewal();
+};
+
+export default renewals;
```
cdmo, line 7: Are you planning on asyncing the "getting" of checkouts later then? I see that the display of checkouts is still syncronously done.

app/javascript/submission_handling/renewal.js
```diff
@@ -0,0 +1,35 @@
...
+import { renderData } from './polling'
+
+const updateCheckout = function (data) {
+    document.querySelector(`[id="checkout${data.item_key}"] .renewal_count`).innerHTML = data.renewal_count;
+    document.querySelector(`[id="checkout${data.item_key}"] .due_date`).innerHTML = data.due_date;
+    document.querySelector(`[id="checkout${data.item_key}"] .status`).innerHTML = data.status;
+};
+
+// This is the public method
+let renewal = function () {
```
cdmo, line 11:
```suggestion
let renewCheckout = function () {
```

app/views/checkouts/_checkouts_table.html.erb
```diff
@@ -16,23 +16,23 @@
...
             <td>
               <%= render BibItemComponent, bibitem: checkout %>
             </td>
-            <td><%= checkout.renewal_count if checkout.renewal_count.positive? %></td>
-            <td><%= render_checkout_due_date checkout %></td>
-            <td><%= render_checkout_status checkout %></td>
-            <td><%= number_to_currency checkout.accrued if checkout.accrued.positive? %>
-            </td>
+            <td class="renewal_count"><%= checkout.renewal_count if checkout.renewal_count.positive? %></td>
+            <td class="due_date"><%= checkout.due_date_human %></td>
```
cdmo, line 33: so this is where you could render something like `"#{checkout.recall_status} <br>" if checkout.recalled?` or something

app/javascript/view_renewals.js
cdmo: I see. This is going to be in https://github.com/psu-libraries/myaccount/issues/196, I forgot about that separate issue.

app/jobs/renewal_job.rb
banukutlu: Yes,I was waiting on the PR to be merged, now that it is I will update :)

app/jobs/renewal_job.rb
banukutlu: We were not showing any specific error msg for these case, was only showing the generic top msg when we were doing the flashes. But will be handled differently with Toast for sure and yeah probably good idea to add a msg here already. Thanks, I will add one.

It is not only overdue necessarily, can be other weird cases. This was one of the sirsi webservices bug. And yes all the overdue ones hit this bug.

app/jobs/renewal_job.rb
banukutlu: I will add a generic msg error for that 👍

app/views/checkouts/_checkouts_table.html.erb
banukutlu: Yes, if recalled, there will be more than one date listed in the due date column because it will add recall due date.

app/jobs/renewal_job.rb
cdmo: ok thanks. I was looking at the holds works i did and wondering if I should change that to be the whole raw error message really too, not sure the best approach

app/jobs/renewal_job.rb
banukutlu: yes, totally redundant, thanks for catching it, forgot to remove those after moving out of the controller :)

app/views/checkouts/_checkouts_table.html.erb
cdmo: ok so in that case I think the method in the model can just be an array of strings then, with some prepended as "Recalled" 👍 (rather than one string with some break tags as separators)

app/jobs/renewal_job.rb
banukutlu: get rid of error_code method all together

app/javascript/submission_handling/renewal.js
banukutlu: hmmm, I was thinking to use checkouts for display checkouts async js 🤔 maybe `renew_checkouts` for this and `view_checkouts` for display?

app/javascript/submission_handling/renewal.js
banukutlu: 👍  will do!

app/javascript/submission_handling/renewal.js
cdmo: yep I like that. I was talking about the filename to be clear... should it still be `renewal.js` or `renew_checkouts.js`? The latter right?

banukutlu: yup, the filename, I added a commit for the change, could you check if looks good?

app/models/checkout.rb
banukutlu: refactored this and moved that logic to a view component, pls let me know what you think about it?

app/jobs/renewal_job.rb
banukutlu: updated

app/jobs/renewal_job.rb
banukutlu: hmm yeah maybe, depending on if holds error messages is sth for user to see, for renewals they are.

cdmo: Well I was thinking of separating the two: one nice message to users and one raw response from sirsi for the logs which would help us debug quicker.

app/models/checkout.rb
cdmo: I was thinking simpler I think, something like this directly in the erb:

```rb
<td>
  <%= "Recalled #{@checkout.recall_due_date} <br>" if @checkout.recalled? %>
  <%= @checkout.due_date %>
</td>
```

I don't think you need to use a view component here because the logic isn't too bad, but maybe I'm missing something

app/javascript/submission_handling/renewal.js
cdmo: thanks that makes more sense to me

cdmo:
Are you just adding display checkouts here too then? Maybe if there isn't going to be a spec we should just put that part in a separate PR? Or are you adding specs for that here too?

banukutlu:
> Are you just adding display checkouts here too then? Maybe if there isn't going to be a spec we should just put that part in a separate PR? Or are you adding specs for that here too?

I will add the specs here as well, could not get to them earlier but working on it.

app/models/checkout.rb
banukutlu: I remember why I had moved the logic to model in the first place, so I do not have to redo it in js when I get the new due date back from the job. It is among the fields that gets updated with renewal. View component is nice but not helping much in this case. I am thinking to move it back to model 🤔

cdmo: It is in the model though, right? You can render this small bit of ERB in the Job itself (server side render it) and store the results in Redis. Maybe we can look at this together over Zoom.

app/jobs/renewal_job.rb
banukutlu: sounds good to me 👍

app/models/checkout.rb
banukutlu: not in the model anymore, I moved it out to the view component, newer version of the model here: https://github.com/psu-libraries/myaccount/pull/195/files#diff-2f2015ebe9d7db97c62891921e1d685c

banukutlu: I am trying to find a way that I can use for display checkout and renew checkout

app/jobs/renewal_job.rb
cdmo: Thanks, looks great

app/jobs/renewal_job.rb
cdmo: Looks good!

app/jobs/renewal_job.rb
cdmo: We can deal with this problem in the toasts issue 👍

app/jobs/renewal_job.rb
cdmo: Cool. So you'll just need to update what is logged to be the raw response from sirsi then right? Currently it is still showing the translation

app/models/checkout.rb
cdmo: I know about the view component.

Why can't you put [this](https://github.com/psu-libraries/myaccount/pull/195#discussion_r427724959) in a partial and render it in a job? If we are producing HTML it should be done by a view not a model is what I'm saying.

.rubocop.yml
```diff
@@ -30,6 +30,7 @@ RSpec/ExampleLength:
   Exclude:
     - 'spec/services/symphony_client_spec.rb'
     - 'spec/features/holds_spec.rb'
+    - 'spec/features/checkouts_spec.rb'
```
cdmo, line 33:
```suggestion
    - 'spec/features/**/*'
```

spec/jobs/renewal_job_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
```
cdmo, line 1: Filename needs to be `renew_checkout_job_spec.rb` right?

spec/controllers/checkouts_controller_spec.rb
```diff
@@ -19,37 +19,70 @@
...
+      it 'requires list of checkouts to be renewed as params' do
+        patch :batch_update, params: {}
+
+        expect(flash[:notice]).to match(/No items were selected/)
+      end
+
+      it 'sends a job to RenewalJob' do
+        patch :batch_update, params: { renewal_list: ['123', '456'] }
+
+        expect(RenewCheckoutJob).to have_received(:perform_later).at_least(2)
```
cdmo, line 76:
```suggestion
        expect(RenewCheckoutJob).to have_received(:perform_later).twice
```

Should be only twice right? Not 2 or more?

banukutlu: yes, twice only, because `renewal_list: ['123', '456']`

spec/jobs/renewal_job_spec.rb
banukutlu: ahhh yes! thanks!

banukutlu: fixed

spec/jobs/renewal_job_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require 'rails_helper'
+
+RSpec.describe RenewCheckoutJob, type: :job do
+  describe '#perform_later' do
+    let(:mock_sc_client) { instance_double(SymphonyClient) }
+    let(:ws_args) { { item_key: 1,
+                      resource: '/catalog/item',
+                      session_token: '1s2fa21465' } }
+    let(:renewal_response) { RENEWAL_RAW_JSON.to_json }
```
cdmo: Have a suggestion for eliminating a duplicative stubbed data: just use sinatra here too like I did on view holds:

```rb
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe RenewCheckoutJob, type: :job do
  describe '#perform_later' do
    let(:ws_args) { { item_key: 1,
                      resource: '/catalog/item',
                      session_token: '1s2fa21465' } }

    before do
      stub_request(:any, /example.com/).to_rack(FakeSymphony)
    end

    after do
      Redis.current.flushall
    end

    context 'with valid input that is returned OK from SymphonyClient' do
      it 'sets a Redis value that marks success' do
        described_class.perform_now(**ws_args)
        results = Redis.current.get 'renewal_1'

        expect(results).to include 'success'
      end

      it 'sets a Redis value that contains a translated item key' do
        described_class.perform_now(**ws_args)
        results = Redis.current.get 'renewal_1'
        renewal_period = 85 # days
        new_due_date_formatted =  (Date.today + renewal_period.days).strftime('%B %-d, %Y')

        expect(results).to eq "{\"item_key\":1,\"result\":\"success\",\"renewal_count\":70,"\
                                      "\"due_date\":\"#{new_due_date_formatted}\",\"status\":null}"
      end
    end
```
And would have to stub an additional bad request to example.com for the failure...

banukutlu: yes, great idea! thanks, I will make the change

app/models/checkout.rb
banukutlu: I added the changes for  due date partial and used it for both renew and display as you suggested

spec/jobs/renewal_job_spec.rb
banukutlu: this is also updated with the last commit

app/jobs/renewal_job.rb
banukutlu: this is also done 👍

spec/support/data/renewal_raw_json.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
```
cdmo, line 1: Sorry, but this can be deleted now right?

spec/jobs/renew_checkout_job_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        stub_request(:post, 'https://example.com/symwsbc/circulation/circRecord/renew?includeFields=circRecord%7B*%7D')
+          .with(body: { item: { resource: '/catalog/item', key: 1 } })
+          .to_return(status: 400, body: error_prompt, headers: {})
+      end
+
+      it 'makes a record of the failure' do
+        described_class.perform_now(**ws_args)
+        results = Redis.current.get 'renewal_1'
+
+        expect(results).to eq '{"item_key":1,"result":"failure","error_message":"Some error message"}'
```
cdmo, line 52: 🆒 I like use Sinatra for this stuff. Feels simpler.

app/jobs/renew_checkout_job.rb
```diff
@@ -0,0 +1,59 @@
...
+      resource: resource,
+      item_key: item_key,
+      session_token: session_token
+    )
+
+    case response.status
+    when 200
+      checkout = Checkout.new(JSON.parse(response.body)&.dig('circRecord'))
+
+      due_date = ApplicationController.renderer.render(
```
cdmo, line 28: Oh didn't know about this approach, interesting.

app/views/checkouts/_due_date.html.erb
```diff
@@ -0,0 +1,3 @@
+<td class="due_date">
```
cdmo, line 1: You know what's kind of weird to me, that upon successful renew this `td` isn't duplicated. It seems to me like you target inner html on this td and replace it with a spinner, then on success it renders the new due date with this partial so there should be a nested td, but there isn't 🤷‍♂️

spec/features/checkouts_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      expect(page).to have_css('[id="checkout3591032:1:1"] .status', text: 'Claims Returned')
+    end
+
+    it 'displays a renew all select' do
+      expect(page).to have_unchecked_field 'renewal_list__2145643:5:1'
+    end
+
+    it 'displays a renew button' do
+      expect(page).to have_button 'Renew'
+    end
```
cdmo, line 84: Part of me wonders if we need all these (or any of these really). Or at least we should think about combining some or all of them because of the test overhead maybe. I'm just not sure what all they provide in terms of assurance in general, and now they're more expensive. 🤔

spec/models/checkout_spec.rb
```diff
@@ -133,4 +125,75 @@
...
+    context 'when an item is claims returned' do
+      it 'returns the right status_human' do
+        checkout.record['fields']['overdue'] = true
+        checkout.record['fields']['claimsReturnedDate'] = '2019-12-01'
+        expect(checkout.status_human).to eq 'Claims Returned'
+      end
+    end
+
+    context 'when an item is overdue' do
+      it 'returns the right status_human' do
```
cdmo, line 139:
```suggestion
      it 'returns "Overdue"' do
```

spec/models/checkout_spec.rb
```diff
@@ -133,4 +125,75 @@
...
+        checkout.record['fields']['overdue'] = true
+        expect(checkout.status_human).to eq 'Overdue'
+      end
+    end
+
+    context 'when an item is not overdue' do
+      it 'returns an empty status_human' do
```
cdmo, line 146:
```suggestion
      it 'is nil' do
```

thinking that "empty" can mean other things

spec/models/checkout_spec.rb
```diff
@@ -133,4 +125,75 @@
...
+
+    context 'when due date time is not 11:59pm' do
+      it 'includes the time' do
+        checkout.record['fields']['recallDueDate'] = '2019-11-09T22:30:00-05:00'
+        expect(checkout.recall_due_date_human).to eq('November  9, 2019 10:30pm')
+      end
+    end
+
+    context 'when an item is not recalled' do
+      it 'returns an empty recall due date' do
```
cdmo, line 192:
```suggestion
      it 'is nil' do
```

spec/support/data/renewal_raw_json.rb
banukutlu: sure, I also removed it from the other raw json files.

app/views/checkouts/_due_date.html.erb
banukutlu: yeah looking at it again now,I agree like how there isn't a nested td there, I do not understand 🤷

spec/features/checkouts_spec.rb
banukutlu: These were in the view spec, I moved them over here because I did not how to test them within a view spec anymore with asyncing work changing thigs 🤔

banukutlu: I think I figure out the view spec, I will move them back to a view spec.

app/views/checkouts/_due_date.html.erb
cdmo: I think we should change it so the surround `td` is outside the partial, who knows, browsers could change this seeming auto-correction fix that's present now.

spec/features/checkouts_spec.rb
cdmo: I know why you moved them here, no I get that. But again, do we need them? What do they actually give us? Take them out and run rake and see what overhead they cost, you know?

app/views/checkouts/_due_date.html.erb
banukutlu: yup, I was just working on changing it :)

spec/features/checkouts_spec.rb
banukutlu: moved them back to a view spec, and cut a few out, there are couple things like displaying Total Overdue, Recalled etc. numbers etc. that I think not covered by other tests. It seems to be not taking much time to run but let me know if you think if these are still an overhead? 

``` 
bundle exec rspec spec/views/checkouts/all.html.erb_spec.rb
...........

Finished in 0.15921 seconds (files took 2.39 seconds to load)
11 examples, 0 failures
```

app/views/checkouts/_due_date.html.erb
banukutlu: updated

spec/features/checkouts_spec.rb
cdmo: much much better. Running all these in a feature spec with `js` on was costing a lot.

PR #209, opened 2020-07-14
Adds feature spec for new hold form population

cdmo:
I also attempted to create a feature spec that shows a successful hold creation (that I didn't include in this PR of course), however it just redirects to example.com because it is trying reauthenticate when the redirect to holds/result happens. Not sure how to get around this, but it'd be nice to have a feature spec for successful hold creation too.

### DIFF
--- a/spec/features/holds_spec.rb
+++ b/spec/features/holds_spec.rb
@@ -49,4 +49,11 @@ RSpec.describe 'Holds', type: :feature do
       expect(page).to have_css '#hold3906718 .hold_status', text: 'Cancelled'
     end
   end
+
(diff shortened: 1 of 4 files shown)


banukutlu:
> I also attempted to create a feature spec that shows a successful hold creation (that I didn't include in this PR of course), however it just redirects to example.com because it is trying reauthenticate when the redirect to holds/result happens. Not sure how to get around this, but it'd be nice to have a feature spec for successful hold creation too.

Thanks for adding a spec for the form and the info here, yes I agree it would be nice to have a spec for hold creation. We can check back this again later.

Issue #167, opened 2020-03-31
Asyncify: New hold requests


cdmo:
In heroku you can select a lot of items (20+) and the browser basically just appears to stop, nothing happens. On the server this happens:

```
2020-03-31T19:01:39.332972+00:00 app[web.1]: I, [2020-03-31T19:01:39.332856 #4]  INFO -- : [06217ac3-ca59-4ab0-9922-6a114c4f5b97] Started POST "/holds" for 71.58.208.129 at 2020-03-31 19:01:39 +0000
2020-03-31T19:01:39.334046+00:00 app[web.1]: I, [2020-03-31T19:01:39.333962 #4]  INFO -- : [06217ac3-ca59-4ab0-9922-6a114c4f5b97] Processing by HoldsController#create as JS
2020-03-31T19:01:39.334140+00:00 app[web.1]: I, [2020-03-31T19:01:39.334086 #4]  INFO -- : [06217ac3-ca59-4ab0-9922-6a114c4f5b97]   Parameters: {"authenticity_token"=>"V7+68tXBzLMvk1KX//SiFxtEkrsGjITL6DW8/82MU7PCfgzRkoFR3PkH6tIcdCjmqSEwSpLy2DXKp7DrJ7rTLw==", "catkey"=>"476022", "pickup_library"=>"UP-PAT", "pickup_by_date"=>"2020-05-31", "barcodes"=>["000059422624", "000059434245", "000059371533", "000059391111", "000059468028", "000059468035", "000059489016", "000059501565", "000059323334", "000061949515", "000061965591", "000061979635", "000061997578", "000062012546", "000062025249", "000062042109", "000062051187", "000062065634", "000062077095", "000059568766"], "commit"=>"Place Hold"}
2020-03-31T19:01:52.672567+00:00 app[web.1]: E, [2020-03-31T19:01:52.672437 #4] ERROR -- : [06217ac3-ca59-4ab0-9922-6a114c4f5b97] Place Hold for 000059422624 by 914261602: {"messageList":[{"code":"hatErrorResponse.722","message":"User already has a hold on this material"}]}
2020-03-31T19:01:52.999344+00:00 app[web.1]: E, [2020-03-31T19:01:52.999181 #4] ERROR -- : [06217ac3-ca59-4ab0-9922-6a114c4f5b97] Place Hold for 000059434245 by 914261602: {"messageList":[{"code":"hatErrorResponse.722","message":"User already has a hold on this material"}]}
[...]

PR #213, opened 2020-07-20
Asyncing new hold requests

cdmo:
* Ports HoldsController#create and HoldsController#results code over to a ActiveJob job
* Creates a new abstraction (model) for SirsiResponse
* For network errors, we alert now
* Removes the Redis setup for session/cache
* Lengthens the duration of polling to 300 seconds. This is mainly due to the problem of requesting multiple items on records with lots of volumetrics.

#167 

This is a draft, can you give a general review? I will spec it out and fix failing specs (there are many) after.


cdmo:
There is an issue of very slow responses for requesting batches of volumes on records with lots of volumes (i.e. try http://localhost:3001/holds/new?catkey=476022). Investigating with @mkutch

app/jobs/place_holds_job.rb
```diff
@@ -0,0 +1,50 @@
...
+      if response.status == 200
+        status[:success] << { barcode: barcode,
+                              hold_key: sirsi_response.hold.key }
+      else
+        Sidekiq.logger.error("place_holds_results_#{patron_key} #{barcode}: #{sirsi_response.response_raw}")
+        status[:error] << { barcode: barcode,
+                            result: :failure,
+                            error_message: sirsi_response.messages.first[:message] }
+      end
+    end
```
cdmo, line 34: I'll add a private method or methods here

cdmo:
This branch needs some changes around checking for busy records.

app/jobs/place_holds_job.rb
banukutlu: yes sounds good! thanks :)

app/services/symphony_client.rb
```diff
@@ -124,7 +124,7 @@ def get_hold_info(hold_key, session_token)
                           }
   end
 
-  def get_item_info(barcode, session_token, key)
+  def get_item_info(barcode = nil, session_token = nil, key = nil)
```
banukutlu, line 127: @cdmo `session_token` should not be nil though, right?

cdmo: so i ran into a problem that i couldnt get around where if I only set key and barcode to nil ruby would throw

app/services/place_hold_results/builder.rb
```diff
@@ -21,18 +21,10 @@ def generate
   private
 
     def hold_lookup(hold_key)
-      parsed_hold = SymphonyClientParser::parsed_response(@client, :get_hold_info, hold_key, @user_token)
-
-      start = DateTime.now
-      while parsed_hold.dig('fields', 'item').nil? && start + 5.seconds > DateTime.now
```
banukutlu, line 27: @cdmo as we discussed, we need a check for item level info

app/models/concerns/sirsi_response.rb
```diff
@@ -0,0 +1,38 @@
...
+    message_array = @response_hash&.dig('messageList')
+    message_array.map do |code_and_message|
+      {
+        code: code_and_message&.dig('code'),
+        message: code_and_message&.dig('message')
+      }
+    end
+  end
+
+  def busy?
```
banukutlu, line 25: liked this abstraction so much 👌  also nice to have this `busy?` method

app/services/place_hold_results/builder.rb
banukutlu, line 27: per our conversation, we need this item level info check for the response, guess we might add to the new SirsiResponse as a method like similar to `busy?`

banukutlu:
> * Ports HoldsController#create and HoldsController#results code over to a ActiveJob job
> * Creates a new abstraction (model) for SirsiResponse
> * For network errors, we alert now
> * Removes the Redis setup for session/cache
> * Lengthens the duration of polling to 300 seconds. This is mainly due to the problem of requesting multiple items on records with lots of volumetrics.
> 
> #167
> 
> This is a draft, can you give a general review? I will spec it out and fix failing specs (there are many) after.

I think overall approach looks great! I liked the sirsi response abstraction very much 👌  As you mentioned, we just need item level info check for the response and adding specs 👍

cdmo:
Thanks Banu, I'll fix things up regarding item level checking and properly busy checking. Then I'll spec it all out. 👍

app/models/concerns/sirsi_response.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class SirsiResponse
+  attr_reader :response_raw, :response_hash
+
+  def initialize(response)
+    @response_raw = response.body
```
banukutlu, line 7: we only need this for error reporting, right?

app/services/symphony_client.rb
banukutlu:
```suggestion
  def get_item_info(session_token, barcode = nil, key = nil)
```

this seems to be working, needs to change in usages as well

app/models/concerns/sirsi_response.rb
cdmo: Like the whole class or just `@response_raw`?

If the latter, I'm only calling it at line 21 of `place_holds_job.rb`

```rb
        Sidekiq.logger.error("place_holds_results_#{patron_key} #{barcode}: #{sirsi_response.response_raw}")
```

You thinking we should get rid of it?

app/services/symphony_client.rb
cdmo: so the order matters then I guess? like it can't be `(barcode = nil, session_token, key = nil)` I know

app/models/concerns/sirsi_response.rb
banukutlu: was asking for `@response_raw`, yeah I just found at `place_holds_job.rb` as well, no we should not get rid of it, I was just trying to understand :)

app/services/place_hold_results/builder.rb
```diff
@@ -25,6 +25,6 @@ def hold_lookup(hold_key)
     end
 
     def item_lookup(barcode)
-      SirsiResponse.new(@client.get_item_info(barcode, @user_token)).item
+      SirsiResponse.new(@client.get_item_info(session_token: @user_token, barcode: barcode)).item
```
banukutlu, line 28: thanks for updating this :)

spec/support/data/other/barcode_not_holdable_barcode.json
```diff
@@ -0,0 +1,8 @@
+{
```
banukutlu, line 1: where do we use this?

cdmo: builder_spec

cdmo: (and fake symphony)

banukutlu: thanks!

app/models/concerns/sirsi_response.rb
cdmo: I wouldn't be opposed :)

app/services/symphony_client.rb
```diff
@@ -195,11 +233,19 @@ def patron_linked_resources_fields(item_details = {})
...
     end
 
+    def hold_info_request
+      authenticated_request "/circulation/holdRecord/key/#{hold_key}",
+                            headers: { 'x-sirs-sessionToken': session_token },
+                            params: {
+                              includeFields: '*,item{*,bib{shadowed,title,author},call{*}}'
+                            }
+    end
+
```
banukutlu, line 243:
```suggestion
```

I could not find where this method is used? maybe I am missing it?

cdmo: no it isn't sorry and thanks

spec/features/holds_spec.rb
```diff
@@ -53,18 +53,12 @@
     end
 
     it 'allows user to place a hold for a holdable item', js: true do
-      inject_into_session place_hold_results: { success: [{
```
banukutlu, line 56: 👍  nice that we do not need this anymore 🎉

spec/jobs/place_holds_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    Redis.current.flushall
+  end
+
+  context 'with valid input' do
+    before do
+      stub_request(:any, /example.com/).to_rack(FakeSymphony)
+    end
+
+    it 'sets a Redis record containing success denoted by patron\'s key' do
+      described_class.perform_now(ws_args)
```
banukutlu, line 24:
```suggestion
      described_class.perform_now(**ws_args)
```

app/jobs/application_job.rb
```diff
@@ -6,4 +6,10 @@ class ApplicationJob < ActiveJob::Base
 
   # Most jobs are safe to ignore if the underlying records are no longer available
   # discard_on ActiveJob::DeserializationError
+  #
+  private
+
+    def symphony_client
+      @symphony_client || SymphonyClient.new
```
banukutlu, line 13: 👌

app/jobs/place_holds_job.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        sirsi_response = SirsiResponse.new response
+
+        if response.status == 200
+          status[:success] << { barcode: barcode,
+                                hold_key: sirsi_response.hold.key }
+        else
+          Sidekiq.logger.error("place_holds_results_#{@patron_key} #{barcode}: #{sirsi_response.response_raw}")
+          status[:error] << { barcode: barcode,
+                              result: :failure,
```
banukutlu, line 53: do we need `result: :failure,` here? Sorry, asking to remember after giving a week long break 😬

spec/jobs/place_holds_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      expect(results).to be_present
+    end
+
+    it 'renders HTML containing the checkouts and saves to redis' do
+      described_class.perform_now(**ws_args)
+      results = Redis.current.get 'place_holds_results_patron1'
+
+      expect(results).to include '1 Hold Placed'
+    end
```
banukutlu, line 35: do we need the first spec? isn't the latter also testing for that case? 🤔  I might be missing sth as well.

app/jobs/place_holds_job.rb
cdmo: In cases of failure, the polling.js script looks for this key: https://github.com/psu-libraries/myaccount/pull/213/files/b3e2a69a16f07f66d661d72b27fc63a4a36f516f#diff-1f1fb8e7901b4fddb481654b8038b693L33

cdmo: Yeah, last you looked at this class it was basically a big long list of instructions, hopefully a bit easier to read with stuff broken down into methods.

spec/jobs/place_holds_job_spec.rb
cdmo: I think the idea is sort of like testing baby steps, but yes, the seconds spec implicitly covers the first. But there could be a scenario in which the first spec passes and the second doesn't. As in, something is wrong with rendering piece but not necessarily the Redis piece. I also was just following the pattern we used in other Job specs, like https://github.com/psu-libraries/myaccount/blob/master/spec/jobs/view_checkouts_job_spec.rb#L18-L29

app/jobs/place_holds_job.rb
banukutlu: ahh thanks, remembered now 👍

spec/jobs/place_holds_job_spec.rb
banukutlu: For the case when second fails and first passes, it tells us yeah this exists but not much other than that, right? I am OK to leave them but I think they are not adding much.

cdmo: In the case of the first spec passing it is a simple test of `Redis.current.set` worked and was able to use `@patron.key` (which in itself is a SymphonyClient call and mapping to a new instance of Patron). So, I think that is a fair amount worth testing, I don't know.

Then aside, we want to make sure this part worked https://github.com/psu-libraries/myaccount/pull/213/files?file-filters%5B%5D=.rb#diff-0bb303924e2704cb7e53481994a126e6R25-R33. Which is some more complexity - service classes and views called on. 

I wouldn't mind keeping both specs and I wouldn't mind deleting the first spec either, but if we do that, we should delete it from the others too, right?

banukutlu: right, if we delete this we should delete the others, I kinda like that but I am also getting eager to merge this PR as well to move forward. It is all good, we can think about them later if we need to.

PR #219, opened 2020-07-28
215 actionview deprecation and upgrade

cdmo:
#215 

After #218 

Also updated the README to remove dev caching instruction which is not needed anymore

Issue #102, opened 2020-01-31
Polish select all component


cdmo:
Like make the label text and placing consistent in holds and checkouts and make it... good.

See https://github.com/psu-libraries/myaccount/issues/83#event-2994823971

cdmo:
Hi @ruthtillman would you mind letting me know your thoughts on options for refactoring the display/behavior of select all:

One option is to remove the first column called "Select" and merge it with "Item" and then have a checkbox next to Item in the table header that can select all items. This would require an aria-label to ensure that it would make sense to unsighted users. 

See https://material-ui.com/components/tables/#sorting-amp-selecting

Another option is to do what is done in elibrary, essentially an empty row that is just for the "All" checkbox.

![Screen Shot 2020-07-29 at 1 12 32 PM](https://user-images.githubusercontent.com/523381/88831516-73bc1c80-d19d-11ea-8cc2-6546f5d47046.png)

Another option might be to make companion "all" buttons. Sort of like below, but for every hold and renew button.

![Screen Shot 2020-07-29 at 1 16 17 PM](https://user-images.githubusercontent.com/523381/88831715-be3d9900-d19d-11ea-9f13-fb859a6de069.png)

I'm not sure which I like best to tell the truth. But I think it's the companion "all" buttons OR the merging of select and item columns.

Maybe a final option is to leave the select alls looking like they do now and say it's good enough.

ruthtillman:
I also like Options 1 and 3. I think that users probably do expect a "select all" option, so perhaps Option 1? Would the checkbox clearly be for all items for sighted users as well? I'm thinking that sometimes without appropriate textual labeling such boxes aren't entirely clear.

cdmo:
It is always tricky to walk the line between "simplifying the interface" and "providing enough context" for best usability IMO. I'll go with option 1 and try to get some options for how that will look. Thanks

PR #225, opened 2020-07-30
102 select all

cdmo:
#102 

after: #219 

Removes visible label from "Select all" option. Also:

* "De-flattens" the buttons
* Change queue position column to be a simple ordinal value.
* Updates Sidekiq gem to the latest


banukutlu:
@cdmo I'm ready to review this PR, rebase pls :)

app/helpers/holds_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module HoldsHelper
   def render_queue_position(hold)
-    return "Your position in the holds queue: #{hold.queue_position}" unless hold.queue_position.nil?
+    return hold.queue_position.ordinalize unless hold.queue_position.nil?
```
banukutlu, line 5: way better like this 👍

Issue #230, opened 2020-08-10
Browser back button leads to stall out


cdmo:
@mkutch reported late last week: when you click the browsers back button to either the checkouts or holds page it leads to a timeout/stall

cdmo:
marking this fixed and closing

PR #231, opened 2020-08-10
force checkouts and holds pages reload on browser back button

banukutlu:
Issue #230

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -37,8 +37,8 @@ class Patron
 
   def checkouts
     @checkouts ||= fields['circRecordList']
-      .map { |checkout| Checkout.new(checkout) }
-      .select { |checkout| checkout.status == 'ACTIVE' }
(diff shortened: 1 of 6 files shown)


app/models/patron.rb
```diff
@@ -37,8 +37,8 @@ def display_name
 
   def checkouts
     @checkouts ||= fields['circRecordList']
-      .map { |checkout| Checkout.new(checkout) }
-      .select { |checkout| checkout.status == 'ACTIVE' }
+                     &.map { |checkout| Checkout.new(checkout) }
+                     &.select { |checkout| checkout.status == 'ACTIVE' }
```
cdmo, line 41: Curious why you added safe navigation here? Does rails throw if there are no checkouts?

banukutlu: One of the specs I added were failing because checkouts were empty for that case I was testing.

cdmo: Love the safe navigation operator, so convenient

Issue #228, opened 2020-08-05
Bills: Title for 'Miscellaneous charges'


banukutlu:
Add `NOT ASSOCIATED WITH AN ITEM` as Title for `Miscellaneous charges`

eg. 

```
[1] pry(#<BibItemComponent>)> bibitem
=> #<Fine:0x00007fdd88427170
 @record=
  {"resource"=>"/circulation/block",
   "key"=>"18511:93",
   "fields"=>
    {"patron"=>{"resource"=>"/user/patron", "key"=>"18511"},
     "amount"=>{"currencyCode"=>"USD", "amount"=>"1.00"},
     "item"=>nil,
     "tax"=>{"currencyCode"=>"USD", "amount"=>"0.00"},
     "billDate"=>"2020-04-29",
     "estimatedOverdueAmount"=>{"currencyCode"=>"USD", "amount"=>"0.00"},
     "title"=>nil,
     "owed"=>{"currencyCode"=>"USD", "amount"=>"1.00"},
     "library"=>{"resource"=>"/policy/library", "key"=>"UP-PAT"},
     "callNumber"=>nil,
     "block"=>{"resource"=>"/policy/block", "key"=>"MISC"},
     "comment"=>nil,
     "itemLibrary"=>nil,
     "createDate"=>"2020-04-29"}}>
```

banukutlu:
@mkutch Could you please check legacy cat myaccount for the title if it is a customization and if so which rule is used?

mkutch:
It's not a customization, but out of the box. If there is no title for the bill then the text "NOT ASSOCIATED WITH AN ITEM" is displayed.

ruthtillman:
Looks good.

PR #232, opened 2020-08-10
customized title for bills with no associated item

banukutlu:
Issue #228 

## Additional Work
Added summary text for no materials and no bills accounts

### DIFF
--- a/app/models/fine.rb
+++ b/app/models/fine.rb
@@ -38,7 +38,7 @@ class Fine
     EQUPREPAIR: 'Fee for repair costs of equipment damaged by borrowers'
   }.with_indifferent_access.freeze
 
-  NO_TITLE_TEXT = 'NOT ASSOCIATED WITH AN ITEM'
+  NO_TITLE_TEXT = 'Not associated with an item'
 
   def initialize(record)
     @record = record
--- a/app/models/fine.rb
+++ b/app/models/fine.rb
@@ -38,8 +38,14 @@ class Fine
     EQUPREPAIR: 'Fee for repair costs of equipment damaged by borrowers'
   }.with_indifferent_access.freeze
 
+  NO_TITLE_TEXT = 'NOT ASSOCIATED WITH AN ITEM'
+
   def initialize(record)
     @record = record
+  end
+
+  def title
+    bib['title'] || NO_TITLE_TEXT
   end
 
   def owed_amount
--- a/app/views/fines/_fines_table.html.erb
+++ b/app/views/fines/_fines_table.html.erb
@@ -14,9 +14,7 @@
   <tbody>
   <% @fines.each do |fine| %>
     <tr>
-      <td>
-        <%= (render BibItemComponent.new(bibitem: fine) unless fine.catkey.nil?) || fine.title %>
-      </td>
+      <td><%= fine_title_display(fine: fine) %></td>
       <td><%= fine.author %></td>
       <td><%= fine.status_human %></td>
       <td><%= number_to_currency fine.owed_amount %></td>
--- a/app/views/fines/_fines_table.html.erb
+++ b/app/views/fines/_fines_table.html.erb
@@ -15,7 +15,7 @@
   <% @fines.each do |fine| %>
     <tr>
       <td>
-        <%= render BibItemComponent.new bibitem: fine %>
+        <%= (render BibItemComponent.new(bibitem: fine) unless fine.catkey.nil?) || fine.title %>
       </td>
       <td><%= fine.author %></td>
       <td><%= fine.status_human %></td>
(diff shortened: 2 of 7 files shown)


app/models/fine.rb
```diff
@@ -38,10 +38,16 @@ class Fine
     EQUPREPAIR: 'Fee for repair costs of equipment damaged by borrowers'
   }.with_indifferent_access.freeze
 
+  NO_TITLE_TEXT = 'NOT ASSOCIATED WITH AN ITEM'
```
cdmo, line 41: Why are we shouting this? 😆

app/models/fine.rb
```diff
@@ -38,10 +38,16 @@ class Fine
...
+
   def initialize(record)
     @record = record
   end
 
+  def title
+    bib['title'] || NO_TITLE_TEXT
```
cdmo, line 48: Is `"block"=>{"resource"=>"/policy/block", "key"=>"MISC"},` more reliable? Or, not having a title is just as good right?

app/views/fines/_fines_table.html.erb
```diff
@@ -15,7 +15,7 @@
   <% @fines.each do |fine| %>
     <tr>
       <td>
-        <%= render BibItemComponent.new bibitem: fine %>
+        <%= (render BibItemComponent.new(bibitem: fine) unless fine.catkey.nil?) || fine.title %>
```
cdmo, line 18: I wonder if it'd be better to remove the `render?` method from `BibItemComponent` and instead build this logic of just printing the title by itself if there is no catkey (from the component itself)?

app/models/fine.rb
banukutlu: The rule they use in current my account is for bills when there is no title so I tried to mimic that.

app/views/fines/_fines_table.html.erb
banukutlu: Two things: We have a `render?` method in the component that validates on catkey so if there is no catkey it does not show which is the case for bills like misc charges. Also this is only for bills afaik (maybe it is not, I should check with Maryam on this) then we will need to check for that as well. I thought about changing the component too but kinda ended up with thinking not to complicate the component further since it is used by checkouts, holds and fines. But yes it can be done like that as well, which would need some logic changes and spec updates.

app/models/fine.rb
cdmo: okie dokie

app/views/fines/_fines_table.html.erb
cdmo: Yeah it's just a little awkward as is to my eyes. Like `catkey` is getting checked for `nil` twice. I get the problem, like rendering an empty BibItemComponent is truthy.

banukutlu: There is no item(bib) present in this case so it kinda made sense not to use the BibItemComponent. 
But if this custom title is also for checkouts and holds as well then it is one less thing to check in the component so I think we can think about changing the `render?` logic there.

spec/models/fine_spec.rb
```diff
@@ -24,4 +24,9 @@
     fine.record['fields']['billDate'] = '2019-10-08'
     expect(fine.bill_date.strftime('%m/%d/%Y')).to eq '10/08/2019'
   end
+
+  it 'has a customized title for Miscellaneous Charges' do
+    fine.record['fields']['item'] = nil
+    expect(fine.title).to eq 'NOT ASSOCIATED WITH AN ITEM'
```
banukutlu, line 30:
```suggestion
    expect(fine.title).to eq 'Not associated with an item'
```

spec/views/fines/index.html.erb_spec.rb
```diff
@@ -37,6 +37,12 @@
       expect(rendered).to include 'A wonderful title'
     end
 
+    it 'displays customized title if the fine does not have an item associated with it' do
+      fine.record['fields']['item'] = nil
+      render
+      expect(rendered).to include 'NOT ASSOCIATED WITH AN ITEM'
```
banukutlu, line 43:
```suggestion
      expect(rendered).to include 'Not associated with an item'
```

app/views/fines/_fines_table.html.erb
cdmo: I see, yes that makes sense. It's an awkward problem for sure.

banukutlu: It is an odd edge case for bills only. Maryam confirmed that it is only for bills, I am inclined toward not much changing the component but what are your thoughts?

cdmo: I wish rendering an empty view component would produce a nil so as to avoid having to double check that catkey is nil. What do you think of a helper to contain all this like 


```rb
def fine_title_display(fine: fine)
  render BibItemComponent.new(bibitem: fine) unless fine.catkey.nil?
  fine.title
end
```

I don't know I don't like this a lot more, but I think it is always better when there is less logic in the partial. 🤷 thoughts? If you feel good about this I'll approve/merge through.

banukutlu: yeah adding a helper is a good idea, will do it.

PR #234, opened 2020-08-14
More details on operating overrides

cdmo:


README.md
```diff
@@ -92,4 +92,6 @@ You can use either the yml file inheritance structure inherent to the config gem
 
 ## Overriding pickup location labels
 
-Pickup locations for holds placed are manually dictated by Lending and Reserves Services. They tell us what they want for the labels and we add them in `settings.yml`. This means we can override them as needed too by following the inheritance flow of the `config` gem. Note that `Settings.pickup_locations` does _not_ affect the labels used in displaying the "Pickup at" column in the holds tables. That is currently not overrideable. 
\ No newline at end of file
+Pickup locations for holds placed are manually dictated by Lending and Reserves Services. They tell us what they want for the labels and we add them in `settings.yml`. This means we can override them as needed too by following the inheritance flow of the `config` gem. Note that `Settings.pickup_locations` does _not_ affect the labels used in displaying the "Pickup at" column in the holds tables. That is currently not overrideable.
```
banukutlu, line 95: which file we will use to override, `settings.yml` or one of the .local.yml?

cdmo: I would say we should never edit any file in our codebase that we track via git. And settings.local.yml is no good because it will just be overridden by deployment. So the only option is `production.local.yml` unless we add our own custom source.

banukutlu, line 95:
```suggestion
Pickup locations for holds placed are manually dictated by Lending and Reserves Services. They tell us what they want for the labels and we add them in `settings.yml`. This means we can override them as needed too by following the inheritance flow of the `config` gem. For production we use `production.local.yml` to override these values. Note that `Settings.pickup_locations` does _not_ affect the labels used in displaying the "Pickup at" column in the holds tables. That is currently not overridable. 
```

Issue #220, opened 2020-07-29
50x template did not appear 


cdmo:
@ruthtillman reported during checkin meeting that making a request to place a hold for a record that does not exist gave a generic browser 50x error

banukutlu:
handled by #240

ruthtillman:
Tested placing hold on record which does not exist. Got appropriate redirect/message.

Issue #237, opened 2020-08-26
holds/new form throws an error when record has no holdables items


banukutlu:
try `/holds/new?catkey=19200019`

<img width="1376" alt="Screen Shot 2020-08-25 at 4 21 32 PM" src="https://user-images.githubusercontent.com/6477867/91349040-b67c0f00-e7b2-11ea-8836-7866689d3fcc.png">

banukutlu:
some notes from debugging this record:

currentLocations on this catkey are CHECKEDOUT, STACKS-BD, STACKS-HB3
STACKS-BD, STACKS-HB3 are not returning back from holdable_locations even though this catkey has IWantIt and availability returns `holdable`

banukutlu:
Update: these locations were updated as holdable but cache needed to be cleared and server needed to be restarted on the sirsi side, after Maryam doing that, now those two locations shows up holdable. 

Also, added a fix for it to not error out in this case.

cdmo:
I'm going to reopen this because I don't think it passed. Please correct me if I'm wrong.

ruthtillman:
I QAd this one. I got the item itself to work successfully but when attempting other items which shouldn't be put on hold at all (just as an edge case), it didn't do anything. Should redirect to my acct home I believe.

cdmo:
Ok cool. Just wasn't sure. Thanks.

banukutlu:
reopened to add a message `Sorry! You cannot place a hold on this item. Contact your library if you need assistance` for records that are not holdable eg. catkey: `31200756`

ruthtillman:
Tested with both online item and item which doesn't exist. Successful!

PR #240, opened 2020-08-31
 fix for new hold form for non holdable records

banukutlu:
Issue #237, also fixes Issue #220

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -4,6 +4,7 @@ class HoldsController < ApplicationController
   before_action :set_cache_headers
   before_action :authenticate_user!
   before_action :check_for_blanks!, only: :create
+  rescue_from NewHoldException, with: :deny_new
   rescue_from HoldCreateException, with: :deny_create
   rescue_from HoldException, with: :past_date
 
@@ -46,6 +47,8 @@ class HoldsController < ApplicationController
                                               user_token: current_user.session_token,
                                               client: symphony_client)
     @place_hold_form_params = form_builder.generate
+
+    raise NewHoldException, 'Error' if @place_hold_form_params.blank?
   end
 
   # Handles placing holds
@@ -86,6 +89,12 @@ class HoldsController < ApplicationController
       return unless barcodes.blank? || params['pickup_library'].blank? || params['pickup_by_date'].blank?
 
       raise HoldCreateException, 'Error'
+    end
+
+    def deny_new
+      flash[:error] = t 'myaccount.hold.new_hold.error_html'
+
+      redirect_to summaries_path
     end
 
     def deny_create
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -43,6 +43,8 @@ class HoldsController < ApplicationController
   #
   # GET /holds/new
   def new
+    raise NewHoldException, 'Error' if params[:catkey].blank?
+
     form_builder = PlaceHoldForm::Builder.new(catkey: params[:catkey],
                                               user_token: current_user.session_token,
                                               client: symphony_client)
@@ -92,7 +94,11 @@ class HoldsController < ApplicationController
     end
 
     def deny_new
-      flash[:error] = t 'myaccount.hold.new_hold.error_html'
+      flash[:error] = if params['catkey'].blank?
+                        t 'myaccount.hold.new_hold.catkey_missing'
+                      else
+                        t 'myaccount.hold.new_hold.error_html'
+                      end
 
       redirect_to summaries_path
     end
--- a/app/services/place_hold_form/builder.rb
+++ b/app/services/place_hold_form/builder.rb
@@ -27,7 +27,7 @@ class PlaceHoldForm::Builder
       title: bib_info.title,
       author: bib_info.author,
       volumetric_calls: @volumetric_calls,
-      barcode: @volumetric_calls.present? ? nil : @call_list&.sample&.items&.sample&.barcode
+      barcode: find_barcode
     }
   end
 
@@ -101,5 +101,9 @@ class PlaceHoldForm::Builder
 
     def volumetric?
       @call_list&.any? { |call| call.volumetric.present? }
+    end
+
+    def find_barcode
+      @volumetric_calls.present? ? nil : @call_list&.sample&.items&.sample&.barcode
     end
 end
--- a/app/services/place_hold_form/builder.rb
+++ b/app/services/place_hold_form/builder.rb
@@ -11,14 +11,7 @@ class PlaceHoldForm::Builder
 
   def generate
     bib_info = Bib.new(SymphonyClientParser::parsed_response(@client, :get_bib_info, @catkey, @user_token))
-
-    if bib_info.call_list.present?
-      @call_list = bib_info.call_list.map { |call| Call.new record: call }
-      filter_holdables if @call_list.count > 1
-    end
-
-    # no holdable items found
-    return {} if @call_list.blank?
+    return {} unless holdables_present?(bib_info)
 
     process_volumetric_calls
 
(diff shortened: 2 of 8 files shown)


cdmo:
[from issue](https://github.com/psu-libraries/myaccount/issues/237#issuecomment-683832671)

> reopened to add a message Sorry! You cannot place a hold on this item. Contact your library if you need assistance for records that are not holdable eg. catkey: 31200756

I just ran http://localhost:3001/holds/new?catkey=31200756 and it did not work but I got the following:

Screen Shot 2020-09-02 at 9 34 09 PM

I was expecting to get "Sorry! You cannot place a hold on this item. Contact your library if you need assistance"

banukutlu:
> [from issue](https://github.com/psu-libraries/myaccount/issues/237#issuecomment-683832671)
> 
> > reopened to add a message Sorry! You cannot place a hold on this item. Contact your library if you need assistance for records that are not holdable eg. catkey: 31200756
> 
> I just ran http://localhost:3001/holds/new?catkey=31200756 and it did not work but I got the following:
> 
> Screen Shot 2020-09-02 at 9 34 09 PM
> 
> I was expecting to get "Sorry! You cannot place a hold on this item. Contact your library if you need assistance"

good catch! this is an interesting one, it is an online resource but has bibinfo and callList. Wonder what should be the criteria to determine if it is holdable or not for these records 🤔

ruthtillman:
While nice to fix if we get a handle on criteria, this is not an issue we need to spend time on. The failure message as screenshotted is sufficient.

cdmo:
Can you give me the catkeys of records you tested with so I can check them out? The reason I caught `31200756` is because it was the one mentioned in the issue. Thanks!

banukutlu:
> Can you give me the catkeys of records you tested with so I can check them out? The reason I caught `31200756` is because it was the one mentioned in the issue. Thanks!

There were couple more that I could not find out right now, will add here if I do:
- 19200019
- 31200756 
- a non-existing key

ruthtillman:
I also used

`2308333`

ruthtillman:
Yields error: To place a hold, please select a volume. `17735134`

banukutlu:
> Yields error: To place a hold, please select a volume. `17735134`

It has a callList but no items with holdable locations so it cannot find a barcode. I think for this it should not load the new form and displays that message, right?

ruthtillman:
Yes, ideally it wouldn't load the form.

app/controllers/holds_controller.rb
```diff
@@ -42,10 +43,14 @@ def batch_update
...
   # GET /holds/new
   def new
+    raise NewHoldException, 'Error' if params[:catkey].blank?
+
     form_builder = PlaceHoldForm::Builder.new(catkey: params[:catkey],
                                               user_token: current_user.session_token,
                                               client: symphony_client)
     @place_hold_form_params = form_builder.generate
+
+    raise NewHoldException, 'Error' if @place_hold_form_params.blank?
```
cdmo, line 53: What is the scenario where this raises that the raise on line 46 doesn't catch?

cdmo (commented):
* `19200019` has holdable items and I was able to add a request for it just fine. 
* `31200756` (a record that just points to a free and open access resource) produced the error "Hold not allowed". 
* a non-existing key produced a Rails flash error of "Sorry! You cannot place a hold on this item. Contact your library if you need assistance."
* no catkey provided produced a Rails flash error of "Please select an item to place a hold."
* `2308333` (unholdable item, needs to be requested through special form) produced "Hold not allowed" like the OA link
* `177351341` (thesis) indeed now does not load the form and instead produces a Rails flash error of "Sorry! You cannot place a hold on this item. Contact your library if you need assistance."


So...

1. Seems like we should maybe use the "Sorry you can't place a hold on this item!" message before allowing the patron to fill out the new hold form for OA links (`31200756`) and special request items (`2308333`) too. Maybe all unholdable items should have this same setup?
1. The error message "Please select an item to place a hold." for when no catkey is specified, I wonder if we should change the language? Because when you get that error you don't have an option to "select" any item to place a hold. It's just kinda confusing to me. I sorta wonder if we even need this at all. I can't imagine this scenario ever happening in the real use of the application? If we want to keep it maybe just make a general error message like "There was a problem processing your hold request, please contact your library if you need assistance." 🤷‍♂️

ruthtillman:
1. Nice to have. Consistency would be ideal. I am ok with launching before that part is consistent because neither of those has an "I Want It" button and therefore we're dealing with an edge case that is already being handled by suppressing "I Want It." So we can prioritize other things before addressing thing.

2. Yes, change to "There was a problem processing your hold request, please contact your library if you need assistance." Ideally they won't run into that either, but if they are we should also know about it and learn how they got there.

cdmo (changes requested):
Ok, so we want to update the language on the no catkey provided error

app/controllers/holds_controller.rb
banukutlu: the first one is for when there is no catkey passed to the new form like by just going to `/holds/new` 
the other one is when the record is not holdable, form_builder will return nil

cdmo: right! :) thanks

config/locales/en.yml
```diff
@@ -68,4 +68,7 @@ en:
...
           success_html: <small class="text-secondary">For placed hold(s), you will receive a library notice once items are available for pick up. On-shelf items generally arrive within 2-3 weekdays. Items checked out to other borrowers will be recalled. Recalled items take a minimum of 10 days to arrive at your pickup location.</small>
-          error_html: <small class="text-secondary">We could not process your request. Please visit our website for more information under <a href="https://libraries.psu.edu/services/borrow-renew#QuestionsReLibraryAcct">"Why did my hold fail?"</a></small>
\ No newline at end of file
+          error_html: <small class="text-secondary">We could not process your request. Please visit our website for more information under <a href="https://libraries.psu.edu/services/borrow-renew#QuestionsReLibraryAcct">"Why did my hold fail?"</a></small>
+        new_hold:
+          catkey_missing: There was a problem processing your hold request, please contact your library if you need assistance.
+          error_html: <span class="font-weight-bold">Sorry!</span> You cannot place a hold on this item. Contact your library if you need assistance.<br>
```
cdmo, line 74:
```suggestion
          error_html: <strong>Sorry!</strong> You cannot place a hold on this item. Contact your library if you need assistance.<br>
```

app/services/place_hold_form/builder.rb
```diff
@@ -11,9 +11,16 @@ def initialize(catkey:, user_token:, client:)
 
   def generate
     bib_info = Bib.new(SymphonyClientParser::parsed_response(@client, :get_bib_info, @catkey, @user_token))
+
+    # record with empty callList
+    return if bib_info.call_list.blank?
```
cdmo, line 16: Can you tell me why there wouldn't be a callList? I am trying to figure out when that scenario occurs. I just checked localhost:3001/holds/new?catkey=31200756 and even that OA electronic resource has a `callList`... 🤔

cdmo (changes requested):
As per our zoom, move filter_holdables invocation into the generate method and add a blank? check on call_list after (just a single blank? check)

it is irritating that `Bib` can have a `@record` of just an error. Like it should not be a `Bib` it should be a `SirsiError` or something. Setting this aside.

app/services/place_hold_form/builder.rb
banukutlu: we talked about this on zoom but just to note here: had to check the callList here because for requests that returns an error, Bib still had a @record of an error

cdmo: yeah 😒

app/services/place_hold_form/builder.rb
```diff
@@ -95,4 +102,8 @@ def naturalize(value)
     def volumetric?
       @call_list&.any? { |call| call.volumetric.present? }
     end
+
+    def find_barcode
```
cdmo, line 106: 👍

cdmo (changes requested):
Overall I think we're good, I'm pushing one commit with some semantics suggestions and code organizing suggestions - can you review?

app/services/place_hold_form/builder.rb
```diff
@@ -11,14 +11,7 @@ def initialize(catkey:, user_token:, client:)
...
     bib_info = Bib.new(SymphonyClientParser::parsed_response(@client, :get_bib_info, @catkey, @user_token))
-
-    if bib_info.call_list.present?
-      @call_list = bib_info.call_list.map { |call| Call.new record: call }
-      filter_holdables if @call_list.count > 1
-    end
-
-    # no holdable items found
-    return {} if @call_list.blank?
+    return {} unless holdables_present?(bib_info)
```
banukutlu, line 14: liked this 👌  good to have the generate method shorter as a summary

spec/services/place_hold_form/builder_spec.rb
```diff
@@ -36,20 +36,20 @@
       expect(form_params[:author]).to eq 'Hill Street blues (Television program)'
     end
 
-    context 'when item is not holdable' do
+    context 'when response from Sirsi is an error, meaning no callList set' do
```
banukutlu, line 39: yes, this is more accurate 👍

banukutlu:
> Overall I think we're good, I'm pushing one commit with some semantics suggestions and code organizing suggestions - can you review?

Your refactor looks good 👌  thanks!

app/services/place_hold_form/builder.rb
cdmo: cool

cdmo (approved):
Teamwork!

banukutlu:
> Teamwork!

🍻

Issue #236, opened 2020-08-26
Redirect to the original request url after auth


banukutlu:
after webaccess, it does not redirect to the original request url, only redirects to summaries

cdmo:
this is in qa

ruthtillman:
QAd! Tested several different kinds of materials, including ones which couldn't be put on hold and everything behaved as expected.

ruthtillman:
Did some more testing this afternoon. It works as expected for first login. However, if one has to re-login, it doesn't pass on the original request. Steps to replicate this issue:

1. On VPN, open un-logged-in browser and place hold, e.g. on https://blackcat01qa.libraries.psu.edu/catalog/25085728. This should work after WebAccess auth.
2. Disconnect VPN. Do something in WebAccess Auth which prompts reauth.
3. Reconnect VPN, in same browser attempt to place a hold on another item, e.g. https://blackcat01qa.libraries.psu.edu/catalog/9357690  
4. You will get WebAccess prompt to reauthenticate and this time it will load homepage.

Maybe this has to do with a slightly different method of reauth vs. auth? Hoping the same solution applies.

banukutlu:
> 1. 25085728

I could not recreate this

banukutlu:
> Did some more testing this afternoon. It works as expected for first login. However, if one has to re-login, it doesn't pass on the original request. Steps to replicate this issue (maybe...some time passed as well, so basically when back on VPN, it did the password re-entry prompt without requiring Duo or anything):
> 
> 1. On VPN, open un-logged-in browser and place hold, e.g. on https://blackcat01qa.libraries.psu.edu/catalog/25085728. This should work after WebAccess auth.
> 2. Disconnect VPN. Do something in WebAccess Auth which prompts reauth.
> 3. Reconnect VPN, in same browser attempt to place a hold on another item, e.g. https://blackcat01qa.libraries.psu.edu/catalog/9357690
> 4. You will get WebAccess prompt to reauthenticate and this time it will load homepage.
> 
> Maybe this has to do with a slightly different method of reauth vs. auth? Hoping the same solution applies.

I was not able to recreate this, it redirected to the correct page 🤔

ruthtillman:
🤔 I will try to see if I can make it happen again this morning...

cdmo:
I experienced a problem in the redirect this morning. When I had to do the webaccess "reauth" thing (not the full DUO authentication) it redirected to the summaries page instead of the new hold request form.

[...]

PR #242, opened 2020-09-02
redirect to original request url

banukutlu:
Issue #236

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,6 +37,8 @@ class ApplicationController < ActionController::Base
     end
 
     def authenticate_user!
+      session[:original_fullpath] = request.original_fullpath unless request.original_fullpath == '/'
+
       return redirect_to root_url unless current_user?
 
       renew_session_token unless symphony_client.ping?(current_user)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,11 +37,15 @@ class ApplicationController < ActionController::Base
     end
 
     def authenticate_user!
-      session[:original_fullpath] = request.original_fullpath unless request.original_fullpath == '/'
+      set_original_fullpath
 
       return redirect_to root_url unless current_user?
 
       renew_session_token unless symphony_client.ping?(current_user)
+    end
+
+    def set_original_fullpath
+      session[:original_fullpath] = request.original_fullpath unless request.original_fullpath == '/'
     end
 
     def renew_session_token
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -37,6 +37,8 @@ def patron_info_response
     end
 
     def authenticate_user!
+      session[:original_fullpath] = request.original_fullpath unless request.original_fullpath == '/'
```
cdmo, line 40: 😎  https://api.rubyonrails.org/classes/ActionDispatch/Request.html#method-i-original_fullpath

spec/controllers/sessions_controller_spec.rb
```diff
@@ -22,6 +22,16 @@
       it 'redirects to the home page' do
         expect(get(:index)).to redirect_to summaries_url
       end
+
+      context 'when user requested a page before authentication' do
```
cdmo, line 26: I tried a fresh browser request of localhost/holds and it redirected to the homepage after authentication. Is that expected behavior? It seems, if I'm reading the logic correctly that unless my request is for the root it should send me to my requested for destination. Right?

cdmo (commented):
Just messing around with this, I gave a shot to just altering `ApplicationController::authenticate_webaccess` to be the following:

```rb
  def authenticate_webaccess
    redirect_to Settings.symws.webaccess_url + request.original_url
  end
```

And it seems to do the trick for all requests and I think it's the only change needed? Did you try this and run into something?

spec/controllers/sessions_controller_spec.rb
banukutlu: right, It should be redirecting to `/holds`. you are reading correctly, not sure why it did not work.

banukutlu:
> Just messing around with this, I gave a shot to just altering `ApplicationController::authenticate_webaccess` to be the following:
> 
> ```ruby
>   def authenticate_webaccess
>     redirect_to Settings.symws.webaccess_url + request.original_url
>   end
> ```
> 
> And it seems to do the trick for all requests and I think it's the only change needed? Did you try this and run into something?

No, I haven't tried this one. It looks nicer with less changes 👍

banukutlu:
> ```ruby
>  + request.original_url
> ```

Just tried it but I could not get it to work. Should we chat/zoom tomorrow for this PR?

spec/controllers/sessions_controller_spec.rb
cdmo: we good

cdmo (changes requested):
feature spec, as per zoom session chat

cdmo:
the feature spec would have been ok I think, but what you have is shorter and clearer 👍

cdmo (approved):

PR #249, opened 2020-09-09
Makes sure we don't present single checkbox during new hold request…

cdmo:
… processing

also a bit of refactoring to improve readability


banukutlu (approved):
👍

PR #250, opened 2020-09-10
Forces renew session actions to go through same auth pipeline as a br…

cdmo:
…and new user (to the application)

I did force a renew by making `renew_session_token` run and it did redirect properly. In master it does not. Basically I forced it to run by assigning the unless to a var, adding a binding.pry and then running the renew unless that var. Then I just tampered with the var by setting it to false when I wanted it to renew.

```rb
      x = symphony_client.ping?(current_user)
      binding.pry
      renew_session_token unless x
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,21 +37,28 @@ class ApplicationController < ActionController::Base
     end
 
     def authenticate_user!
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
seems to be working 👍  thanks

PR #251, opened 2020-09-11
Refines datadog settings

cdmo:
I deployed this up to QA, really the only way to debug.

You can see all 'myaccount' prefixed services here: https://app.datadoghq.com/apm/services?env=production&paused=false&search=myaccount


cdmo:
Screen Shot 2020-09-11 at 12 11 32 PM

VERSION
```diff
@@ -0,0 +1 @@
+v0.3.2
```
banukutlu, line 1: 👌

cdmo: kinda of a pain in a way because now we have to keep this in mind when doing PRs and deployments... idk. We don't need this but I thought it might be a nice to have in the datadog view? Then again I'm not sure. 🤷

banukutlu: yeah, let's give it a time to see if we use it for datadog then worth it 👍

banukutlu (approved):
looks good 👍

Issue #247, opened 2020-09-09
Accessibility Test Fixes Required


cdmo:
- [x] remove "--" in pickup locations select
- [x] Run deque axe matchers and fix all the things

cdmo:
In the screen reader testing we have a glaring error of nothing being announced to the screen after renew or hold update. I am fixing that and it’s good. But… there is no definitive announcement like “successfully renewed” and Banu and I were thinking some subtle notification to sighted users might be helpful too. So, we came up with the idea of using a badge under the item to indicate things went well and this will also get announced to screenreaders. There may be a better more elegant solution out there, but I think we do need something..

<img width="979" alt="Screen Shot 2020-09-10 at 2 04 14 PM" src="https://user-images.githubusercontent.com/523381/92778938-9ed68600-f36f-11ea-8fd3-6c71c966a350.png">

(more on “badges” from twitter bootstrap: https://getbootstrap.com/docs/4.5/components/badge/) 

Ruth likes this idea too 👍

ruthtillman:
Badges look good.

PR #253, opened 2020-09-14
247 a11y

cdmo:
#247 

Just to show work in progress on what is starting to become a larger overhaul of the application for a11y reasons.

### DIFF
--- a/app/components/select_all_child_component.rb
+++ b/app/components/select_all_child_component.rb
@@ -4,10 +4,11 @@ class SelectAllChildComponent < ViewComponent::Base
   def initialize(name: nil, key:, target_keyword:)
     @name = name || "#{target_keyword}_list[]"
     @key = key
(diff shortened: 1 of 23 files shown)


cdmo:
This PR adds more in the way of ARIA role attributes which helps partially solve the no announcement problem, but I will plan to handle a clearer success indication notification in a future PR (#255). I think the work done so far is a clean stopping point

app/components/select_all_child_component.rb
```diff
@@ -4,10 +4,11 @@ class SelectAllChildComponent < ViewComponent::Base
   def initialize(name: nil, key:, target_keyword:)
     @name = name || "#{target_keyword}_list[]"
     @key = key
+    @id = "#{@name}[#{key}]"
```
banukutlu, line 7: 👌

banukutlu (approved):
👍

Issue #255, opened 2020-09-14
Success notifications


cdmo:
- [x] the application doesn't notify user of successful renew or hold edit/cancel
- [x] badge for success indicator
- [x] scroll up on success

from #253

ruthtillman:
Tested and looks good. A note on timing -- the scroll up happens immediately and then one may scroll down and see that items are still in the process of renewing. This isn't a criticism, just a timing note. I think it works because one knows something happened when the page scrolls up, the one scrolls down to see status. Items which can be renewed get nice green flare, items which can't get red popups. 👍

PR #257, opened 2020-09-15
255 success notification

cdmo:
for #255 

~this will need a rebase after #253~ rebased


banukutlu (approved):
looks great! very liked these badges 👌 kinda even would like to see those for failed actions :)

PR #260, opened 2020-09-16
Experimental: A more reliable, less manual intervention method of getting the curre…

cdmo:
this ok you think? just trying to trim down the machinery we have to think about


spec/support/fake_webservice.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'sinatra/base'
+
+class FakeWebservice < Sinatra::Base
```
banukutlu, line 5: like it better moved over to here 👍

spec/support/fake_github.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'support/fake_webservice'
+
+class FakeGithub < FakeWebservice
```
banukutlu, line 6: 👍

banukutlu (approved):
looks good to me 👌

PR #261, opened 2020-09-16
make spinner smaller

banukutlu:
also show only one spinner per checkout for renewals

Note: Need to get Ruth's feedback on this change.

### DIFF
--- a/app/javascript/submission_handling/shared.js
+++ b/app/javascript/submission_handling/shared.js
@@ -9,7 +9,7 @@ export let submitterValue = (event) => {
 
 export const findForm = (form) => document.querySelector(`form#${form}`);
 
-export const spinner = `<div class="spinner-border" role="status">
+export const spinner = `<div class="spinner-border spinner-border-sm text-secondary" role="status">
                             <span class="sr-only">Loading...</span>
                         </div>`;
 
--- a/app/javascript/submission_handling/shared.js
+++ b/app/javascript/submission_handling/shared.js
@@ -23,7 +23,9 @@ export let responseFromRails = (event) => {
     return 'Not defined';
 };
 
-export const toggleSpin = function (cssPath) {
+export const toggleSpin = function (type, key, className) {
+    const cssPath = `[id="${type}${key}"] .${className}`;
+    document.querySelector(`${cssPath} span`).classList.toggle('invisible');
     if (document.querySelector(`${cssPath} .spinner-border`)) {
         document.querySelector(`${cssPath} .spinner-border`).remove();
     } else {
@@ -31,15 +33,6 @@ export const toggleSpin = function (cssPath) {
         if (document.querySelector(`${cssPath} .warning`)) {
             document.querySelector(`${cssPath} .warning`).remove();
         }
-    }
-};
-
-// eslint-disable-next-line max-params
-export const toggleResults = function (type, key, className, showSpinner = true) {
-    const cssPath = `[id="${type}${key}"] .${className}`;
-    document.querySelector(`${cssPath} span`).classList.toggle('invisible');
-    if (showSpinner) {
-        toggleSpin(cssPath)
     }
 };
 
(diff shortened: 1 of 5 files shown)


app/javascript/submission_handling/shared.js
```diff
@@ -36,6 +34,15 @@ export const toggleSpin = function (type, key, className) {
     }
 };
 
+// eslint-disable-next-line max-params
+export const toggleResults = function (type, key, className, showSpinner = true) {
```
cdmo, line 38: Small suggestion: I think I'd prefer to have the emptying out of a table cell severed from the spin toggling, I think they're kinda coupled up here and agree with the linter that we probably should be wary of sending so many parameters. It just becomes kinda complex and hard to read and maybe a sign that the function is doing too much? I'm suggesting a separate function for clearing contents, like:

```js
const clearContents = (type, key, className) => {
    const cssPath = `[id="${type}${key}"] .${className}`;
    document.querySelector(`${cssPath} span`).classList.toggle('invisible');
}
```

Thoughts?

I also kinda wondered why you ditched your first commit where you just emptied out the html directly from `renew_checkout.js`. I thought [this](https://github.com/psu-libraries/myaccount/pull/261/commits/4d1d89a1c3d81af4eebbef67a9268d80a2e34a81#diff-2ae0ea096eaad5177977d7cecca79d90R30) was even better than writing a separate function 🤔

cdmo (changes requested):
see comment in line with code

app/javascript/submission_handling/shared.js
banukutlu: I actually started with that exact function then I thought CC would ask me to refactor because the same lines are being used in both methods idk maybe it would not, that is why I ditched that one. But let me actually commit it that way and see if CC will nag.

I ditched the emptying out because when the renewal fails I need the original value and with the failed response we do not have that info only a messageList, so I went with toggling it.

cdmo: maybe just use the invisible class toggle then whereever you need to either turn it on or off?

banukutlu: isn't that what I am doing? 🤔

banukutlu: yeah I will seperate it from spin and see if CC will be OK but if you mean sth else I did not understand sorry :/

cdmo: I meant just call `toggle('invisible');` in the code rather than wrapping it inside another function. What you are doing is calling a function that wraps that invocation along with alternatively toggling the spinner. I'm saying just toggle the spinner or toggle the invisible class directly from the js like

```js
    # for example
    findForm('checkouts').addEventListener("submit", function () {
            allChecked(findForm('checkouts')).forEach((checkbox) => {
                toggleSpin('checkout', checkbox.value, 'renewal_count');
                document.querySelector(`[id="checkout${checkbox.value}"] .due-date span`).classList.toggle('invisible');
                document.querySelector(`[id="checkout${checkbox.value}"] .status span`).classList.toggle('invisible');
```

Does that make sense?

banukutlu: aahhh I see! funny I had it this way too, I cannot remember why I changed it to combine with the toggle LOL
anyway yes, I will change it this way

banukutlu: guess maybe thinking we may need to toggle visible w/o spinner in other places as well, but yeah currently there are no other.

cdmo: Sorry this has been hard to explain, that's on me! I think toggleSpin can remain how it was and we can just toggle the invisible class directly for the items we don't want to also spin.

So like 

> guess I was thinking that we may need to toggle visible w/o spinner in other places as well

In that case we're already covered, because you can just call `element.classList.toggle('invisible');` - that's the use case I'm talking about. When we need toggle that visibility and show a temp spinner then we can run toggleSpin

Let's zoom to go over it. I'm probably missing something.

banukutlu: I think we are on the same page, you are not missing anything 🤔

cdmo (approved):

PR #262, opened 2020-09-18
Yarn upgrade of webpack-dev-server to bump node-forge

cdmo:
Just to remediate CVE-2020-7720 - not that it's a concern for us because this is dev only.


banukutlu (approved):
👍

PR #264, opened 2020-09-21
revised loading msg

banukutlu:
part of Issue #2


cdmo (approved):
Would be nice to offer some translations of our application in other languages some day

Issue #229, opened 2020-08-07
Consider using policy for bill Reason 


mkutch:

cdmo:
there isn't much to review, backend change. closing

PR #267, opened 2020-09-23
bill reason from policy

banukutlu:
Issue #229


cdmo (approved):
Clean 👍

cdmo:
If it turns out we need a fallback to code option for the status_human of a the fine, we can fixup later

banukutlu:
> If it turns out we need a fallback to code option for the status_human of a the fine, we can fixup later

sounds good

Issue #266, opened 2020-09-23
Success Badges Disappear With New Button Click


ruthtillman:
Success badges for renewals/hold updates/etc., should disappear when the user clicks a new button. Because we don't reload the whole page, they aren't otherwise disappearing.

Sample task for this would be updating hold pickup locations and then hold request by dates.

ruthtillman:
Yes/and -- I have a couple screenshots to review during our meeting.

PR #268, opened 2020-09-23
clear all badges

banukutlu:
Issue #266

### DIFF
--- a/app/javascript/submission_handling/shared.js
+++ b/app/javascript/submission_handling/shared.js
@@ -50,3 +50,8 @@ export const scrollToTop = () => {
         "behavior": 'smooth'
     });
 };
+
+export const clearBadges = () => {
+    const badges = document.getElementsByClassName('badge-success');
+    while (badges.length > 0) badges[0].remove();
+};
--- a/app/javascript/submission_handling/shared.js
+++ b/app/javascript/submission_handling/shared.js
@@ -1,3 +1,5 @@
+const defaultIndex = 0;
+
 export let allChecked = (form) => form.querySelectorAll('.checkbox:checked');
 export let submitterValue = (event) => {
     if (event.submitter) {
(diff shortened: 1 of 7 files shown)


app/javascript/submission_handling/shared.js
```diff
@@ -49,4 +49,9 @@ export const scrollToTop = () => {
         "left": 0,
         "behavior": 'smooth'
     });
+};
+
+export const clearBadges = () => {
+    const badges = document.getElementsByClassName('badge-success');
+    while (badges.length > 0) badges[0].remove();
```
cdmo, line 56: Overall comment: This approach seems fine though a bit tricky, like a race maybe between clearing old badges and clearing new ones by mistake in the case where new badges appear very quickly. 

What about adding a new listener to ajax:before? And in that listener you can trigger a badge clear. See https://guides.rubyonrails.org/working_with_javascript_in_rails.html#rails-ujs-event-handlers

Aside: stylistically I would prefer curly brackets as `js-lint` [is asking for](https://eslint.org/docs/2.0.0/rules/curly) - this is kinda hard to read

banukutlu: @cdmo would you mind doing a another quick review for the changes I added?

cdmo: You didn't use a new listener for ajax:before... ?

banukutlu: I moved the clearBadges calls to existing `ajax:success`listeners for example https://github.com/psu-libraries/myaccount/pull/268/files#diff-c144f09703f116bdd78fd8aee6f84d69R40-R42

cdmo: Ok, and I think that's better, but you didn't use a new listener for ajax:before and my comment was just wondering why. It seemed to me like a handy way to make sure, for sure, that the badges are all cleared out before any of the ajax requests, spinning and value updating occurred. What you have here is probably fine though 👍

banukutlu: clearing before spinning and etc. makes sense, let me change to add ajax:before

cdmo: Ok thanks. It's probably fine how you have it. I just get worried about weird things happening in JavaScript land when it comes to order of DOM manipulation.

cdmo (approved):
👌

PR #270, opened 2020-09-25
Caches the version for 5 minutes and refactors class

cdmo:
I noticed in prod logs that we are getting errors regarding rate limits, so caching for 5 minutes to avoid problems

### DIFF
--- a/app/lib/myaccount_version.rb
+++ b/app/lib/myaccount_version.rb
@@ -1,17 +1,13 @@
 # frozen_string_literal: true
 
 module MyaccountVersion
-  def self.version
-    retrieve_version
+  def self.resolve_version
+    set_version
+    Redis.current.get 'myaccount_version'
   end
 
   VERSION_CACHE_TIME = 300
   VERSION_NOT_SET_NAME = 'Version could not be determined'
-
-  def self.retrieve_version
-    set_version
-    Redis.current.get 'myaccount_version'
-  end
 
   # `nx: true` means that Redis will only set this value if it doesn't exist. The `ex` option tells Redis when to delete
   # the key.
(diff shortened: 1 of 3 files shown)


app/lib/myaccount_version.rb
```diff
@@ -2,17 +2,31 @@
 
 module MyaccountVersion
   def self.version
-    version_from_github || 'not set'
+    retrieve_version
```
banukutlu, line 5: I am sure I am missing sth here, why not `self.version` just does what `self.retrieve_version`, why another method needed 🤔

cdmo: My thinking was to separate them for code clarity and clear responsibility of methods - also to make the one public method simple and the private methods handle the complexity. I wanted to the `version` method to have one task: get the version. So to do that it calls `retrieve_version`. That method has more responsibility of "resolving the version" lets say. Calling `retrieve_version` `resolve_version` might have been a better name. 

Happy to change this if you want.

banukutlu: Is `self.retrieve_version` private? I did not realize that 🤔

My instinct was like to check where else`retrieve_version` is called by but then it is not, idk maybe I am not thinking `retrieve_version` is too complex - totally my perception. Or maybe retrieve method calling `set_version` kinda confuses me, idk. I might be not understanding well, too.

cdmo: Ok, I'll change it. I somehow left out the `private` keyword.  It is meant to go under self.version and above  `VERSION_CACHE_TIME = 300`

banukutlu: thanks!

banukutlu (approved):
thanks for the change, this was easier to understand for me :)

Issue #286, opened 2020-10-15
Back to Catalog link catkey issue


banukutlu:
Back to Catalog link should be linked to catkey

PR #287, opened 2020-10-15
286 place hold catkey

banukutlu:
#286 

Additional: Removed leftover github api specs support stuff


cdmo (approved):
thanks for getting leftover stuff i forgot from GH API work 🙏

Issue #289, opened 2020-10-20
Add a message about quarantined materials


banukutlu:
Add the following line somewhere on the Checkouts or Summary page:

"Returned materials may be quarantined for four days and will remain on your account during the quarantine period."

Maybe bold, italicized or red font to draw attention.

cdmo:
Chris okayed this via email

PR #290, opened 2020-10-20
Adds quarantine notice as a callout

cdmo:
#289 

Also makes the dashboard display at small screen a bit cleaner

Screen Shot 2020-10-20 at 9 37 50 AM
Screen Shot 2020-10-20 at 9 36 22 AM

### DIFF
--- a/app/views/shared/_callout.html.erb
+++ b/app/views/shared/_callout.html.erb
@@ -1,0 +1,8 @@
+<div class="row">
+  <div class="col-md-6">
+    <div class="bs-callout bs-callout-danger">
+      <h3><i class="fa fa-exclamation-circle" aria-hidden="true"></i> Notice</h3>
+      <p>Returned materials may be quarantined for four days and will remain on your account during the quarantine period.</p>
+    </div>
+  </div>
+</div>
--- a/app/views/shared/_callout.html.erb
+++ b/app/views/shared/_callout.html.erb
@@ -1,8 +1,8 @@
 <div class="row">
   <div class="col-md-6">
     <div class="bs-callout bs-callout-danger">
-      <h3><i class="fa fa-exclamation-circle" aria-hidden="true"></i> Notice</h3>
-      <p>Returned materials may be quarantined for four days and will remain on your account during the quarantine period.</p>
+      <h3><%= t 'myaccount.callout.label_html' %></h3>
+      <p><%= t 'myaccount.callout.message' %></p>
     </div>
   </div>
 </div>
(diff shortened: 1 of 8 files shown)


app/views/shared/_callout.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="row">
```
banukutlu, line 1: what do you think about naming this partial sth other than `_callout`, idk what would be a good name maybe like `_covid_notice` or sth like that, `_callout` could be confusing a bit 🤔

cdmo: it's a callout, as in a https://cpratt.co/twitter-bootstrap-callout-css-styles/

might be that we will reuse it in the future for something other than covid is what I was thinking

banukutlu: sure, it is using the callout but we might have other partials that uses callout as well right?

app/views/shared/_callout.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <div class="col-md-6">
+    <div class="bs-callout bs-callout-danger">
+      <h3><i class="fa fa-exclamation-circle" aria-hidden="true"></i> Notice</h3>
+      <p>Returned materials may be quarantined for four days and will remain on your account during the quarantine period.</p>
```
banukutlu, line 5: can the message added to en.yml?

cdmo: I guess 🤷 

Sometimes I feel like adding the i18n component is kind of a waste of time because we literally only have english. I'll move it though that's fine.

app/views/shared/_callout.html.erb
cdmo: Hmmm, like where? I am not thinking of another place where this would go. 

I do think calling it `_covid_notice` is destining it to be renamed or removed.

If we end up using it more than one place I'd reach for a View Component called "callout" I think, but, to me that's not needed yet

app/views/shared/_callout.html.erb
banukutlu: thanks, I think it might make this partial look more generic

app/views/shared/_callout.html.erb
banukutlu: sth like just `_notice` would be fine too, but we can leave it as is 👍

cdmo: I like callout because it's semantic with the UI component fwiw

banukutlu (approved):
look good 👌

Issue #288, opened 2020-10-16
holds/result check for referer


banukutlu:

cdmo:
Think we should close 0.3x and mark the last deploy as 1.0, right?

cdmo:
I'm sorry, but what is the problem that this issue is discussing? I think the title of the issue is the solution, but what was the problem? I apologize because I think this is something that you and Maryam spoke about but I didn't catch it.

banukutlu:
I found about this while checking the issue Maryam brought up, it is again sth that we do not expect as a normal user behavior but if you directly go to /holds/result the page just keeps spinning, this is to check if it is a redirect from from holds/new

cdmo:
So... we are fixing the problem of a weird experience for someone who types in `myaccount.libraries.psu.edu/holds/result` - is that right?

banukutlu:
Yes, a new hold is not placed unwantedly or anything like that, just the page displays this msg `Attempting to place holds, please wait...`, it is just misleading.

cdmo:
Ok yup, I did get it. Before I just didn't wait long enough. Sorry!

ruthtillman:
Looks good!

PR #292, opened 2020-10-20
check referer on holds/result

banukutlu:
#288 

This is probably not a regular user case but just in case if somebody directly goes to `/holds/result`.

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -6,7 +6,6 @@ class HoldsController < ApplicationController
   before_action :check_for_blanks!, only: :create
   rescue_from NewHoldException, with: :deny_new
   rescue_from HoldCreateException, with: :deny_create
-  rescue_from HoldException, with: :past_date
 
   # Render patron holds
   #
@@ -75,7 +74,7 @@ class HoldsController < ApplicationController
     if request.referer && URI(request.referer).path == '/holds/new'
       render
     else
-      redirect_to summaries_path
+      redirect_to '/not_found'
     end
   end
 
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -72,7 +72,11 @@ class HoldsController < ApplicationController
   #
   # GET /holds/result
   def result
-    render
+    if request.referer && URI(request.referer).path == '/holds/new'
+      render
+    else
+      redirect_to summaries_path
+    end
   end
 
   # Handles form submission for canceling holds in Symphony
--- a/spec/controllers/holds_controller_spec.rb
+++ b/spec/controllers/holds_controller_spec.rb
@@ -182,10 +182,30 @@ RSpec.describe HoldsController, type: :controller do
     end
 
     describe '#result' do
-      it 'renders the result template' do
-        get :result
+      context 'when not redirected from /holds/new' do
+        it 'redirects to the summary page' do
+          get :result
 
-        expect(response).to render_template(:result)
+          expect(response).to redirect_to summaries_path
+        end
+      end
+
+      context 'when redirected from /holds/new' do
+        let(:place_hold_params) {
+          { catkey: '1',
+            barcodes: ['1'],
+            pickup_library: 'UP_PAT',
+            pickup_by_date: '2050-02-02' }
+        }
+
+        before do
+          allow(PlaceHoldsJob).to receive(:perform_later)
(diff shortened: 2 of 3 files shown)


spec/controllers/holds_controller_spec.rb
```diff
@@ -189,10 +189,30 @@
...
+      context 'when not redirected from /holds/new' do
+        it 'redirects to the summary page' do
+          get :result
 
-        expect(response).to render_template(:result)
+          expect(response).to redirect_to summaries_path
+        end
+      end
+
+      context 'when redirected from /holds/new' do
```
banukutlu, line 200: this might be redundant because we are testing the same thing for create
since this is a controller spec, cannot use sth like follow_redirect! to test the rendered template
that can be done with a request spec though, not sure if we need that.

let me know your thoughts..

cdmo (commented):
So, thinking about this problem of "I went to /holds/result directly" some more. The problem is that it spins a long time and nothing happens. In general I think this is the smallest of edge case scenarios, but I believe the user would eventually realize their mistake and move on (it would have to be a mistake on their part and not intentional).

Given that we want to "fix" this I wonder if there might not be a better option thanredirecting to the summaries page immediately. That seems kinda odd to me. I think I'd rather see something like an error message like a [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) and some indication to the user that the page is "stale." This can be done via parameter check I think.

If having the route /holds/result is a problem, another or related solution could be to remove the route and instead of redirection to a result page in the Rails code you could just load the results in place on the same page as the form using [a small amount of javascript](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#remote-elements).

I think I would like to see both a 400 bad request check as well as removing the route in favor of an ajax-ed form result to fix this problem..

banukutlu:
> So, thinking about this problem of "I went to /holds/result directly" some more. The problem is that it spins a long time and nothing happens. In general I think this is the smallest of edge case scenarios, but I believe the user would eventually realize their mistake and move on (it would have to be a mistake on their part and not intentional).

redis job eventually fails with an alert `There was a network error, please try again later or call your campus library.` 

> 
> Given that we want to "fix" this I wonder if there might not be a better option thanredirecting to the summaries page immediately. That seems kinda odd to me. I think I'd rather see something like an error message like a [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) and some indication to the user that the page is "stale." This can be done via parameter check I think.
> 
> If having the route /holds/result is a problem, another or related solution could be to remove the route and instead of redirection to a result page in the Rails code you could just load the results in place on the same page as the form using [a small amount of javascript](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#remote-elements).
> 
> I think I would like to see both a 400 bad request check as well as removing the route in favor of an ajax-ed form result to fix this problem..

I rather not get into a ajax-ed fix esp. for this very edge case, throwing a 400 bad request and displaying a msg sounds good.

cdmo:
I never got that "network error" message when I tried. I think I waiting 4 or 5 minutes...

banukutlu:
> I never got that "network error" message when I tried. I think I waited 4 or 5 minutes...

Interesting, I think I always get it, this screenshot is from prod but I get it on local and qa as well, not sure how long it took but yeah we have a `const maxWaitTime = 300000;` 5 min. max time
 
Screen Shot 2020-10-22 at 10 56 59 AM

app/controllers/holds_controller.rb
```diff
@@ -6,7 +6,6 @@ class HoldsController < ApplicationController
   before_action :check_for_blanks!, only: :create
   rescue_from NewHoldException, with: :deny_new
   rescue_from HoldCreateException, with: :deny_create
-  rescue_from HoldException, with: :past_date
```
banukutlu, line 9: I think this is not used anymore 🤔  so I removed it

cdmo: Yeah I think it and it's model can be deleted, good eyes!

spec/controllers/holds_controller_spec.rb
cdmo: Yeah I would get rid of this second spec because it's redundant as you say.

banukutlu: ok, will do, wanted to see what you think 👍

cdmo (approved):

Issue #295, opened 2020-10-21
Renewing items with session and active login that is expired


mkutch:
I had a myaccount open from the previous day. I clicked on Renew items and it's just spinning.  I was able to recreate it in Chrome when I login in to my account and then cleared browsing data.  Same thing happens with any Hold activities too. Perhaps there should be some kind of redirect back to the catalog after certain amount of time, or check if the patron session has timed out and it redirects them back to catalog?

cdmo:
That is supposed to be happening. Thanks for reporting.

cdmo:
@mkutch you made this issue at 11:18 am this morning, can you tell me roughly when you attempted to renew and it just spun?

cdmo:
@mkutch can you test this out on QA? Maybe leave your browser open over night there and see what happens tomorrow morning? Also, though it's not the exact same scenario, delete the website's cookie in-session to test too?

cdmo:
this is in prod now, will close as we discussed in check-in

PR #299, opened 2020-10-23
Adds an error condition and reaction for AJAX requests

cdmo:
#295 

Handles `422` server responses by quickly displaying a message that session is stale

Screen Shot 2020-10-22 at 11 28 44 PM

I attempted a feature spec, tried a few variations on:

```rb
# checkouts_spec.rb
context 'when patron\'s session has become stale' do
    it 'shows a modal telling the user to start a new session', js: true do
      page.driver.browser.manage.delete_all_cookies
      page.check 'renewal_list__2145643:5:1'
      page.click_button 'Renew', match: :first
      sleep 3
      expect(page).to have_css '#expiryModal'
    end
  end
```

but not quite there. Might be impossible to test?

### DIFF
--- a/app/controllers/checkouts_controller.rb
+++ b/app/controllers/checkouts_controller.rb
@@ -19,9 +19,9 @@ class CheckoutsController < ApplicationController
     render
   end
 
-  # Handles form submission for renewals in Symphony
+  # Handles form submission for batch checkout renewal requests
   #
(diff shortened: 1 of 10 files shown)


cdmo:
added some feature specs to emulate the bad token situation

banukutlu (approved):
looks great 👍

Issue #274, opened 2020-10-06
Support users whose default location does not allow pickups during closure period


ruthtillman:
Because some libraries aren't doing pickup, it's possible a user's default location will not be allowing pickups.

If their default location is not matched in the list _when placing holds_, then make it blank to force selection.

banukutlu:
@mkutch could we find a test account with a default library that is not doing pickup right now?

ruthtillman:
Successfully tested with my default library, UP, removed from QA list! Looks good to go.

PR #310, opened 2020-10-28
If default location is not in pickup locations list, add a blank option to force selection

banukutlu:
#274 

Additionally, a change to check filter_holdables even if there is only one holding - this only covers withdrawn records and such and users could not get to those records anyway but I think it does not hurt to change the check to cover it.

### DIFF
--- a/app/services/place_hold_form/builder.rb
+++ b/app/services/place_hold_form/builder.rb
@@ -35,7 +35,7 @@ class PlaceHoldForm::Builder
       return false if bib_info.call_list.blank?
 
       @call_list = bib_info.call_list.map { |call| Call.new record: call }
-      filter_holdables if @call_list.count > 1
+      filter_holdables if @call_list.count.positive?
 
       @call_list.present?
     end
--- a/spec/helpers/holds_helper_spec.rb
+++ b/spec/helpers/holds_helper_spec.rb
@@ -158,5 +158,10 @@ RSpec.describe HoldsHelper, type: :helper do
       expect(helper.render_pickup_libraries('UP-PAT')).to include '<option selected="selected" value="UP-PAT">'\
                                                                           'Pattee Commons Services Desk</option>'
     end
+
+    it 'returns an html select list with a blank option as selected' do
+      expect(helper.render_pickup_libraries('not-a-pickup-location')).to include '<option selected="selected" '\
+                                            'disabled="disabled" value="">Please choose a pickup location</option>'
+    end
   end
 end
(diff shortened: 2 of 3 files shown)


spec/helpers/holds_helper_spec.rb
```diff
@@ -158,5 +158,10 @@
       expect(helper.render_pickup_libraries('UP-PAT')).to include '<option selected="selected" value="UP-PAT">'\
                                                                           'Pattee Commons Services Desk</option>'
     end
+
+    it 'returns an html select list with a blank option as selected' do
+      expect(helper.render_pickup_libraries('not-a-pickup-location')).to include '<option selected="selected" '\
+                                            'disabled="disabled" value="">Please choose a pickup location</option>'
+    end
```
cdmo, line 165: maybe add a spec to cover the `.blank?` context?

app/services/place_hold_form/builder.rb
```diff
@@ -35,7 +35,7 @@ def holdables_present?(bib_info)
       return false if bib_info.call_list.blank?
 
       @call_list = bib_info.call_list.map { |call| Call.new record: call }
-      filter_holdables if @call_list.count > 1
+      filter_holdables if @call_list.count.positive?
```
cdmo, line 38: This would _not_ solve the problem you encountered yesterday though, right? Because the call list was empty on that made-up catkey execution would never get past the guard statement at the top. 

This change is for a different scenario. It is challenging the assumption that when there is a call list there is at least 1 holdable call. Is that right?

cdmo (changes requested):
Would like to see a spec for the `.blank?` and have a question about the change to `builder.rb`

app/services/place_hold_form/builder.rb
banukutlu: It solves the problem I had yesterday, eg. catkey: 473854, this record is withdrawn and call list is not empty, has 1 holding.

spec/helpers/holds_helper_spec.rb
banukutlu: sure, good idea, will do 👍

app/services/place_hold_form/builder.rb
cdmo: ah I thought it had an empty call list

banukutlu: yeah, it has one call with library key `ZREMOVED`

cdmo: ok fair enough. it does seem like a slim chance of happening in-real-life, but guess why not

banukutlu: honestly very unlikely, like maybe if the url is saved and used again later and then in the meantime the record would changed to withdrawn - sounds almost no way but still I thought it would not hurt to check that one holding :shrug:

cdmo (approved):

Issue #265, opened 2020-09-22
Add language re: Illiad/Medianet note appearing in checkouts or renewals


ruthtillman:
For checkouts/renewals page, display following message: "Materials borrowed through Illiad or Medianet do not appear on this list and must be renewed separately."

For holds page, display following: "Materials requested through Illiad or Medianet do not appear on this list and must be managed separately."

ruthtillman:
Text is good -- can we link both "Illiad" and "Medianet" in a way that takes the user through auth to those sites? ... related, I don't know as much about Medianet and if it's still in place and how it works. If y'all don't either, lets check with Binky.

ruthtillman:
@banukutlu After looking at the Medianet link, checked its name w/Binky. She says replace "Medianet" with "MTSS".

ruthtillman:
It looks good except it's only appearing on Holds page vs. Checkouts/Renewals

banukutlu:
> It looks good except it's only appearing on Holds page vs. Checkouts/Renewals

fix added

ruthtillman:
QA completed, looks good.

PR #311, opened 2020-10-28
info message for checkouts and holds

banukutlu:
#265  

Also removed some unused keys from en.yml


cdmo (approved):

cdmo:
I might add a PR to fix existing markup structure in `app/views/checkouts/all.html.erb` - nesting divs and paragraphs inside a paragraph feels really wrong to me. Not part of this PR of course.

PR #314, opened 2020-10-30
attempt for a workaround to back button to holds/result issue

banukutlu:
#312 

This is probably not a good sol'n but the only one I could find, I am open to other ideas, wanna hear your thoughts @cdmo.

Also to note here, checking document.referrer did not work in this case because when back button is used, browser is serving from the cache and returns the original referrer.

### DIFF
--- a/app/javascript/submission_handling/polling.js
+++ b/app/javascript/submission_handling/polling.js
@@ -31,6 +31,7 @@ const validateResult = (data, otherRule) => {
 };
 
 const checkError = (data) => data.result === 'failure';
+const checkIfCached = () => performance.getEntriesByType("navigation")[0].type === "back_forward";
 
 const pollFetch = function(arg, otherRule = null) {
     const maxWaitTime = 300000;
@@ -43,6 +44,8 @@ const pollFetch = function(arg, otherRule = null) {
                 resolve(data);
             } else if (checkError(data)) {
                 resolve(data);
+            } else if (checkIfCached()) {
+                reject(new Error('not found'));
             } else if (Number(new Date()) < endTime) {
                 setTimeout(checkCondition, pollInterval, resolve, reject);
             } else {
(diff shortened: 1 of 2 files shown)


app/javascript/submission_handling/polling.js
```diff
@@ -31,6 +33,7 @@ const validateResult = (data, otherRule) => {
 };
 
 const checkError = (data) => data.result === 'failure';
+const checkIfCached = () => performance.getEntriesByType("navigation")[defaultIndex].type === "back_forward";
```
cdmo, line 36: 😰

banukutlu: :sob:

app/javascript/submission_handling/polling.js
```diff
@@ -2,6 +2,8 @@
 import 'jquery'
 import 'bootstrap'
 
+const defaultIndex = 0;
```
cdmo, line 5: would this "back_forward" value ever be at `1`? or some other key? If not I'd suggest something like "navigationKeyLocation" or whatever?

cdmo (changes requested):
I think this needs a feature spec :/

app/javascript/submission_handling/polling.js
cdmo: I don't know. Maybe just `expectedArrayPosition`? idk

banukutlu: I think it would be 0 in our case, `performance.getEntriesByType("navigation")` returns a list of PerformanceEntry objects for every mark set. https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType. But when no marks set, it seems to contain a single object that has the navigation information for the window. 

Screen Shot 2020-11-03 at 12 26 53 PM

banukutlu:
> I think this needs a feature spec :/

yeah I was thinking about adding one but did not want to do it before we both agree on the sol'n

cdmo:
I am hoping that a larger refactor will eliminate the need for this entirely. But as a short term solution it appears to work, right?

banukutlu:
> I am hoping that a larger refactor will eliminate the need for this entirely. But as a short term solution it appears to work, right?

It worked on my local, would you mind testing on your local as well?

Also https://caniuse.com/nav-timing

cdmo:
I did look it up on caniuse. Had never heard of this API before.

Locally it works. What I did is place a hold, go to my holds, click back. And then I got a page not found.

banukutlu:
> I think this needs a feature spec :/

spec added

app/javascript/submission_handling/polling.js
cdmo: right - so that's what you expect the position of the value to be in the array right? Just a suggestion about a name change as `defaultIndex` to me isn't super clear

cdmo (approved):

app/javascript/submission_handling/polling.js
banukutlu: oh yeah I could have change that, sorry I forgot about this one, will make that change in a another PR

cdmo: it's fine I think

Issue #328, opened 2020-11-17
Explore replacing polling/sidekiq for new holds requests


cdmo:

cdmo:
still waiting QA/stage server connection ability. Can't review until it has a server to ping

banukutlu:
QA'ed and deployed ✔️

PR #329, opened 2020-11-18
Dials down Sidekiq retries to just 1 and adds some documentation.

cdmo:
#328

this is a reversion mainly. New code is in the second commit


banukutlu:
@cdmo this one needs a rebase ;)

banukutlu:
just realized the branch name `polling-returns` LOL

app/controllers/holds_controller.rb
```diff
@@ -132,8 +125,4 @@ def deny_create
     def barcodes
       [params['barcodes']].flatten.compact
     end
-
-    def item_details
-      { holdRecordList: true }
-    end
```
banukutlu, line 138: no need for this anymore? what about the one in checkouts_controller?

app/models/patron.rb
```diff
@@ -2,10 +2,6 @@
 
 # Class to model Patron information
 class Patron
-  include ActiveModel::Validations
-  include ActiveModel::Conversion
-  extend ActiveModel::Naming
-
```
banukutlu, line 8: don't we need these for the patron validation check on key?

app/services/symphony_client.rb
```diff
@@ -43,9 +43,6 @@ def patron_info(patron_key:, session_token:, item_details: {})
                                        ].join(',')
                                      })
     JSON.parse(response.body)
-  rescue HTTP::TimeoutError => e
-    # Allow execution to proceed. Error response handled by validating Patron creation downstream.
-    puts "Rescued: #{e.inspect}"
```
banukutlu, line 48: no need for this anymore?

banukutlu (changes requested):
couple questions re removing patron validation

app/models/patron.rb
cdmo: yes, this was a mistake, should be corrected now

app/controllers/holds_controller.rb
cdmo: 🤔 do we? I wasn't sure tbh, but page loads were fast to the spinner. I'll double check. This happened because a straight reverse in git is what I did of both of the previous PRs.

app/services/symphony_client.rb
cdmo: this was _needed_ for the prior work that was reversed... I don't know that we _need_ it anymore but maybe keeping it would be a good idea. Lemme think about that.

app/controllers/holds_controller.rb
banukutlu: I am not sure either, we are passing it in the job as well. thanks for checking, I am curious.

cdmo: no we can't have this, and I removed the one that was still there for `checkouts_controller.rb`. The reason is that on load of `/checkouts` (or `/holds`), the first thing that happens is `authenticate_user!` - this will in turn look up the current user and make a `Patron` out of it. To make a patron it uses `SymphonyClient.patron_info`. That method gets passed `item_details` on the controller being requested. This method will dictate what parameters are sent to the web service. The default was made to be _only_ `alternateId` so it could be super speedy on first load. If you pass in `{circRecordList: true }` it'll want to get all the checkouts too. 

Checkouts, in this sidekiq approach, are pulled in by a job instead, that job directly sets the requirement for what is needed. (i.e.,`{ circRecordList: true }`.

banukutlu: yes, makes sense 👍  thanks!

app/services/symphony_client.rb
banukutlu: any thoughts on this? not keeping it?

README.md
```diff
@@ -59,6 +59,12 @@ Monitor the behavior by tailing the logs:
  ```
  bundle exec sidekiq
  ```
+ 
+ Sidekiq is being used as a means of speeding things up for our end users. This is especially true for large batches of changes being requested and most true for large batches of changes that apply to a single bib (multiple calls). Without the concurrency that Sidekiq provides, these requests would take significantly more time. For example a request to place a hold on 32 volumes might take 106 seconds without Sidekiq and with Sidekiq that comes down to a still long, but much quicker 15 seconds. 
+ 
+ Another option could have been hijacking the HTML and sending individual HTTP requests from JavaScript and then handling the response on each in JavaScript as well, but, we feel that this is approach is less attractive due to the fact that it's even more complex than the Sidekiq and polling apparatus and seems to work against the intention of a form in HTML rather than with it.
```
banukutlu, line 65:
```suggestion
 Another option could have been hijacking the HTML and sending individual HTTP requests from JavaScript and then handling the response on each in JavaScript as well, but, we feel that this approach is less attractive due to the fact that it's even more complex than the Sidekiq and polling apparatus and seems to work against the intention of a form in HTML rather than with it.
```

app/jobs/application_job.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  sidekiq_options retry: 1
```
banukutlu, line 4: question: why not `retry: 0` or even `retry: false`? won't setting it to 1 let the jobs to be retried once?

cdmo: Yes, just once. My thinking on this was that we "might as well" let it do just one because of "network jitter" potential. There's a chance something just blanks out quickly and comes back quickly. But, I woudn't want it to go more than once I think. Basically thinking like "ok let's just keep retries alive and see what we see" kind of thing. Kind of an experiment. I do think we've handled most of the cases ourselves in terms of looking for problematic responses from the web service, but, maybe there are legit ones that could be solved by a quick retry? I guess I'm just second guessing the idea of complete elimination of retries for now - mainly because it doesn't hurt to try once more, there is no cost to the user and no cost in complexity.

banukutlu: my concern is about the use case we thought while we were discussing to disable the retries. Like for example if the user place tried to update location on a hold but then for some reason the job failed, then user went ahead and tried again was able to change the location but this time choose another location. Then the previous job retry kicked in and changed the location to the previous chose of user's. I guess the scenario was sth like this 🤔 When does the retry happen?

cdmo: Yes, this is a concern for sure. I was thinking that because the first retry happens within only a second or two that would mitigate the issue. So in the above scenario you'd have a user sitting there looking at a spinner waiting for the thing to finish, which would only be a few more seconds tacked on. 

Starting to gravitate back to just turning it off completely again though, who knows how long the initial timeout or whatever uncaught error will take to be produced and then retried on.

app/services/symphony_client.rb
cdmo: nope, not keeping it. probably would have been better to do rescue HTTP::Error as a blanket catch, but regardless, no. a rescue here prevents the desired action from happening - which is a retry. If it times out after a retry it'll just fail and after a time the javascript will prompt the user that a bad thing happened. I was looking for a way to sort of detect that we were on the last attempt of the job and then rescue there and return back an error, but I couldn't find a way to do that in ActiveJob. Sidekiq has a method, but overriding the time between tries formula seemed too difficult and perhaps [even not recommended](https://github.com/mperham/sidekiq/issues/4281#issuecomment-543815196)

app/jobs/application_job.rb
cdmo: changed in latest

app/jobs/application_job.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
-  # Automatically retry jobs that encountered a deadlock
-  # retry_on ActiveRecord::Deadlocked
+  # Using ActiveJob to manage retries, only if there's a HTTP related problem. Using 0 seconds because AJ adds jitter
+  # time. In reality "1 second" can be up to 5 and 0 is around 1-2.
+  retry_on(HTTP::Error, wait: 0.seconds, attempts: 2)
```
banukutlu, line 6: 👌

app/jobs/application_job.rb
```diff
@@ -1,12 +1,12 @@
...
+  # Telling Sidekiq to discard jobs instead of retrying (job failed twice already)
+  sidekiq_options retry: false
```
banukutlu, line 8: 👌

app/services/symphony_client.rb
banukutlu: I agree let's not keep it

banukutlu (approved):
👌

PR #330, opened 2020-11-18
Forces Warden to be cleared in cleanup of specs using login_permanent…

cdmo:
…ly_as

glad we found this problem, not entirely satisfied with the solution

https://github.com/wardencommunity/warden/issues/99#issuecomment-496196068


banukutlu (approved):
thanks for the fix! 👌

PR #332, opened 2020-12-08
Using existing data from request to process renewal

cdmo:


banukutlu (approved):
looks good, one thing for later is that we need to handle the case for when attempting a renewal with no checkouts selected.

Issue #331, opened 2020-12-03
Announcement bar functionality like is in the catalog


cdmo:
like

![Screen Shot 2020-12-03 at 3 58 19 PM](https://user-images.githubusercontent.com/523381/101087588-687c6180-3580-11eb-8599-ed42dd557698.png)

cdmo:
this is in prod now, but @banukutlu  you want to review it?

banukutlu:
> this is in prod now, but @banukutlu you want to review it?

yup, will do, thanks!

banukutlu:
closing this, reviewed on qa and it worked 👍

PR #337, opened 2020-12-10
adds an announcement bar

banukutlu:
#331


cdmo (changes requested):
The default state of this feature is a blank white bar at the top of the page. Is that intentional? Thinking we should either set it as false or provide a default that works without configuration?

banukutlu:
> The default state of this feature is a blank white bar at the top of the page. Is that intentional? Thinking we should either set it as false or provide a default that works without configuration?

I was thinking I had it set as false, my mistake, will fix

cdmo (approved):
Maybe an example in the readme would help too?

banukutlu:
> Maybe an example in the readme would help too?

yes good idea, will do

cdmo (approved):
👌

Issue #335, opened 2020-12-10
500 when "record is in use" is returned by SymphonyClient


cdmo:
[This kind of error](https://app.datadoghq.com/apm/trace/3780943963685493527?spanID=1171035985006238457&env=production&sort=time&colorBy=service&graphType=flamegraph&shouldShowLegend=true) is already being tested for and retried by our own code, we just need to catch it too as it is expected behavior from the web service. Relevant code is https://github.com/psu-libraries/myaccount/blob/master/app/services/symphony_client.rb#L237-L255

screenshot of sidekiq log:

![Screen Shot 2020-12-10 at 9 11 55 AM](https://user-images.githubusercontent.com/523381/101783044-c6043700-3ac7-11eb-8678-62aa8e13b095.png)

Looks like a lot of the errors we're seeing in datadog are related to this problem, [some more](https://splunk.ois.psu.edu/en-US/app/search/search?q=search%20index%3Dlibraries%20host%3Dl5bcapp01prod*%20source%3D%22%2Fvar%2Flog%2Fmessages%22%20process%3Dsidekiq&display.page.search.mode=smart&dispatch.sample_ratio=1&earliest=1607580763&latest=1607580883&sid=1607610539.192267_0AC0DB48-E8E4-45B3-ABBA-58A6EC0DCD8E)

I think this is an "inevitable exception" and that we should "branch around it" but I don't think that's possible. Think we have to use it as part of control flow (reading [How to Rescue Exceptions in Ruby](https://stackify.com/rescue-exceptions-ruby/)). Basically this has to be a part of control flow IMO

PR #341, opened 2021-01-12
datadog ignore records in use

banukutlu:
#335 monkey patching datadog to not report records in use error


banukutlu:
@cdmo on two things I'd like to see what you think: 1. should we ignore other errors as well? 2. should we keep that more details message on the error span?

cdmo (changes requested):
What do you think about putting this code inside a Module that gets called by the datadog initializer? See https://www.justinweiss.com/articles/3-ways-to-monkey-patch-without-making-a-mess/ except we're not monkeypatching a core Ruby lib, rather a third party gem, so it could be like `lib/gem_extensions/class_name/group.rb` instead I'm thinking.

cdmo:
I don't think we should ignore other errors no, this one is definitely the exception, hopefully the lone one 🤞 

I'm a little worried about having datadog consume the message and what that might mean for the app:

> We had a discussion on #1117, and it is not safe to read the http.rb response body in ddtrace: we end up consuming the body reading stream, so the host application cannot properly read it; we could end up with a very large payload in memory (think hypothetically a large download file as the body).

from https://github.com/DataDog/dd-trace-rb/issues/1116#issuecomment-664471091

cdmo:
I was able to see that this code is seeming to work in my test. I did a multi (maybe selected 15 or so) volumes of https://catalog.libraries.psu.edu/catalog/3039887 and I generated the record in use error as seen below:

Screen Shot 2021-01-14 at 9 13 36 AM

And DataDog did not see it. 👍  So it does appear to be working.

I think instead of the `begin ... rescue` we could instead use safe navigation right?

banukutlu:
> I don't think we should ignore other errors no, this one is definitely the exception, hopefully the lone one 🤞
> 
> I'm a little worried about having datadog consume the httprb body message and what that might mean for the app:
> 
> > We had a discussion on #1117, and it is not safe to read the http.rb response body in ddtrace: we end up consuming the body reading stream, so the host application cannot properly read it; we could end up with a very large payload in memory (think hypothetically a large download file as the body).
> 
> from [DataDog/dd-trace-rb#1116 (comment)](https://github.com/DataDog/dd-trace-rb/issues/1116#issuecomment-664471091)

that is the reason I pointed out to see if we wanna keep message like if there is extra value for us and ideally better not to keep it. It was nice to have for testing this patch :)

banukutlu:
> group

that is the link I used as well initially, and tried that way but could not get it to work then I ditched it for an easier not the idal way - reason it did not work which now I am guessing because I was not restarting sidekiq, you know I realized that one way later. I can give it a try again.

cdmo:
shoot, i spoke too soon, the error was reported I think, [see](https://app.datadoghq.com/apm/traces?end=1610636567420&paused=false&query=-service%3A%28etda-workflow-redis%20OR%20etda-workflow%20OR%20etda-workflow-mysql2%20OR%20myaccount-redis%20OR%20myaccount-cache%20OR%20myaccount-sidekiq%20OR%20sidekiq-client%20OR%20myaccount%29%20status%3A%28error%20OR%20ok%29%20env%3Aqa&spanID=589907775714964071&spanViewType=metadata&start=1610622167420&storage=hot&streamTraces=true&trace=AQAAAXcBO9sdg4a6dgAAAABBWGNCUERMekFBQnhoR2R6MnRWdUZ0a2k&traceID=2511294369904803743)

cdmo (approved):

Issue #339, opened 2020-12-16
Add error code 238 to sirsi response translations


banukutlu:
This error code returned when couple users tried to renew Hathi ETAS items which were auto-renewed. We got servicenow tickets (INC1592348, INC1592358) because the users got the generic error message. Adding a custom message would be more clear for user.

```{"messageList"=>[{"code"=>"hatErrorResponse.238", "message"=>"User cannot check out this item"}], "describeUri"=>"/exception/advicePrompt", "promptRequired"=>false, "dataMap"=>{"recommendedAction"=>"/circulation/itemCircInfo/advise"}}```

cdmo:
Thanks for adding this. I wonder if this 238 error also shows up for non-ETAS things.

banukutlu:
> Thanks for adding this. I wonder if this 238 error also shows up for non-ETAS things.

yeah, I assume it would show up but do not know yet in which cases. @mkutch is this sth we can find out?

cdmo:
Just not sure what the translation would or should be, what are you going to make it?

banukutlu:
> Just not sure what the translation would or should be, what are you going to make it?

Not sure yet, after finding out in which cases this code is returned we might be able to come up with sth. and will get @ruthtillman input on the msg text as well.

cdmo:
Ok, cool. Saw you self assigned so was wondering if you were jumping on it or not. Think it's safe to get to in January 👍

banukutlu:
> Ok, cool. Saw you self assigned so was wondering if you were jumping on it or not. Think it's safe to get to in January 👍

yeah, not starting on it yet, will wait ;)

banukutlu:
@ruthtillman any thoughts on what should the custom error msg for this error be? Currently, it does not seem much helpful? @mkutch in which cases do we get `hatErrorResponse.238` other than renewal attempt on HathiETAS items?

ruthtillman:
Talked to Chris, we settled on "This item cannot be automatically renewed. Please contact ul-lending@lists.psu.edu for assistance in renewing the item."

banukutlu:
tested on QA ✔️ 
<img width="366" alt="Screen Shot 2021-02-22 at 3 51 07 PM" src="https://user-images.githubusercontent.com/6477867/108768722-02233180-7526-11eb-9a1a-98852aa706d7.png">

ruthtillman:
Looks great!

cdmo:
deployed

PR #351, opened 2021-01-27
adding error code 238 to sirsi translation

banukutlu:
#339

### DIFF
--- a/app/services/sirsi_response/error.rb
+++ b/app/services/sirsi_response/error.rb
@@ -9,6 +9,8 @@ class SirsiResponse::Error
     "hatErrorResponse.105": 'Denied: Item has been recalled, cannot be renewed.',
     "hatErrorResponse.252": 'Denied: Item has holds, cannot be renewed.',
     "hatErrorResponse.46": 'Denied: Item on reserve, cannot be renewed.',
+    "hatErrorResponse.238": 'This item cannot be automatically renewed. Please contact '\
+          '<a href="mailto:ul-lending@lists.psu.edu">ul-lending@lists.psu.edu</a> for assistance in renewing the item.',
     "unhandledException": 'Denied: Item cannot be renewed.'
   }.with_indifferent_access
 
--- a/app/services/sirsi_response/error.rb
+++ b/app/services/sirsi_response/error.rb
@@ -9,8 +9,10 @@ class SirsiResponse::Error
     "hatErrorResponse.105": 'Denied: Item has been recalled, cannot be renewed.',
     "hatErrorResponse.252": 'Denied: Item has holds, cannot be renewed.',
     "hatErrorResponse.46": 'Denied: Item on reserve, cannot be renewed.',
-    "hatErrorResponse.238": 'This item cannot be automatically renewed. Please contact '\
-          '<a href="mailto:ul-lending@lists.psu.edu">ul-lending@lists.psu.edu</a> for assistance in renewing the item.',
+    "hatErrorResponse.238": ActionController::Base.helpers.sanitize(
+      'This item cannot be automatically renewed. Please contact '\
+      "#{ApplicationController.helpers.mail_to 'ul-lending@lists.psu.edu'} for assistance in renewing the item."
+    ),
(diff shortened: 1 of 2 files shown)


cdmo (changes requested):
I believe (sanitize)[https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize] is meant for scrubbing user input. I think I'd recommend using (`html_safe`)[https://apidock.com/rails/String/html_safe] instead.

Also curious as to how you tested locally. Sinatra?

banukutlu:
> I believe [sanitize](https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize) is meant for scrubbing user input. I think I'd recommend using [`html_safe`](https://apidock.com/rails/String/html_safe) instead.
> 
> Also curious as to how you tested locally. Sinatra?

yes right, I changed it, thanks 👍 

In terms of testing, I think this one kinda covers the translated error msg already https://github.com/psu-libraries/myaccount/blob/main/spec/services/sirsi_response/error_spec.rb#L51. Locally I just tested it with another translated error not necessarily with a 238 but one of the others.

cdmo:
I wasn't talking about specs really, more the "how are you simulating it locally" - I wasn't sure how that worked. Probably just mock the response somehow right

app/services/sirsi_response/error.rb
```diff
@@ -9,6 +9,8 @@ class SirsiResponse::Error
     "hatErrorResponse.105": 'Denied: Item has been recalled, cannot be renewed.',
     "hatErrorResponse.252": 'Denied: Item has holds, cannot be renewed.',
     "hatErrorResponse.46": 'Denied: Item on reserve, cannot be renewed.',
+    "hatErrorResponse.238": 'This item cannot be automatically renewed. Please contact '\
+          '<a href="mailto:ul-lending@lists.psu.edu">ul-lending@lists.psu.edu</a> for assistance in renewing the item.',
```
cdmo, line 13:
```suggestion
          '<a href="mailto:ul-lending@lists.psu.edu">ul-lending@lists.psu.edu</a> for assistance in renewing the item.'.html_safe,
```

This is actually what I was thinking, but, does that not work how I am thinking it does?

cdmo: Separately I wonder why we have double quotes for the keys of this `Hash`. Seems kinda weird considering we never use double quotes for strings without interpolation anywhere else and rubocop usually yells at us for that. 🤷

cdmo: Could also use [`mail_to`](https://apidock.com/rails/v5.2.3/ActionView/Helpers/UrlHelper/mail_to), instead of marking the html as safe, like:

```suggestion
          "mail_to('ul-lending@lists.psu.edu') for assistance in renewing the item.",
```

banukutlu: ah this is even better 👍  thanks

banukutlu: > Separately I wonder why we have double quotes for the keys of this `Hash`. Seems kinda weird considering we never use double quotes for strings without interpolation anywhere else and rubocop usually yells at us for that. 🤷

yes very weirdly rubocop does not yell at this. I will change those quotes

banukutlu: just tried mail_to but did not work. I think it requires helper method to be included in this model.

cdmo: ah sorry "#{mail  to here}"

banukutlu: I'd tried that too :/

cdmo: just adding a note here about convo in slack: basically sidekiq loaded string here does not have the method html_safe and the mail_to helper is undefined it seems which seems to imply that RoR isn't properly loaded/initialized for sidekiq jobs perhaps 🤔

cdmo (approved):

banukutlu:
tested on QA ✔️ 
Screen Shot 2021-02-22 at 3 51 07 PM

Issue #354, opened 2021-02-23
Exclude more sirsi responses from datadog v2


banukutlu:
- Error 500: [{"code"=>"hatErrorResponse.722", "message"=>"User already has a hold on this material"}]
- Error 500: [{"code"=>"hatErrorResponse.17286", "message"=>"Hold not allowed"}]
- {"messageList":[{"code":"hatErrorResponse.753","message":"User already has this title charged out"}]}

banukutlu:
reviewed on QA ✔️  ready for prod

PR #357, opened 2021-02-24
one more dd exclude

banukutlu:
#354 continue'd


banukutlu:
@cdmo sorry found one more while testing on local. there will probably more but thought good to get these out of the way.

cdmo (approved):
image

Issue #352, opened 2021-02-09
"I want it" not working from Summon


mkutch:
Searched for a book in Summon then clicked "I want it" link in Summon. Took me to the Myaccount page with this message "Sorry! You cannot place a hold on this item. Contact your library if you need assistance." I can place a hold on the book from the Catalog or CAT.

cdmo:
I think I see what's happening

I just went to https://psu.summon.serialssolutions.com/advanced#!/search/document?ho=t&l=en&q=(TitleCombined:(walk%20in%20the%20woods))&id=FETCHMERGED-psu_catalog_a179440442 and clicked the "I Want It" link in the right sidebar and experienced this issue. Looking at the url provided by the I Want It link, there seems to be a typo:

`https://myaccount.libraries.psu.edu/holds/new?catkey=a17944044`

It should be

`https://myaccount.libraries.psu.edu/holds/new?catkey=17944044`

So there's an extra little a after the equal sign. @ruthtillman is this something you fix? Or is this a thing you fix @mkutch? Not sure who manages this in Summon...

ruthtillman:
Thanks for the diagnosis, @cdmo -- the problem here is that Summon thinks the catkeys begin with letters, e.g. "a" etc. So we need to find a way to have Summon strip those off or strip off the letter "a" on arriving catkey values. I will investigate Summon's side of things and talk to Maryam.

ruthtillman:
Opened case with Summon: 00924829

banukutlu:
@ruthtillman Just checking to see if there are any updated from Summon on this?

Also we are gonna sniff the leading `a` and remove it for placing holds.

ruthtillman:
No action from Summon yet. I updated the ticket once. Thank you!

PR #359, opened 2021-02-25
fix to delete prefix from catkey for place holds coming from Summon

banukutlu:
#352

### DIFF
--- a/spec/services/place_hold_form/builder_spec.rb
+++ b/spec/services/place_hold_form/builder_spec.rb
@@ -33,6 +33,18 @@ RSpec.describe PlaceHoldForm::Builder do
       expect(builder.generate[:author]).to eq 'Hill Street blues (Television program)'
     end
 
+    context 'when catkey has a prefix "a"' do
+      let(:catkey) { 'a1' }
+
+      before do
+        allow(client).to receive(:get_bib_info).with('1', 'fake-token').and_return(get_bib_info_response)
+      end
+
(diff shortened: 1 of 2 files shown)


spec/services/place_hold_form/builder_spec.rb
```diff
@@ -33,6 +33,18 @@
       expect(builder.generate[:author]).to eq 'Hill Street blues (Television program)'
     end
 
+    context 'when catkey has a prefix "a"' do
+      let(:catkey) { 'a1' }
+
+      before do
+        allow(client).to receive(:get_bib_info).with('1', 'fake-token').and_return(get_bib_info_response)
```
cdmo, line 40: that catkey variable defined on 37 never gets called - shouldn't it be loaded into get_bib_info as the first argument instead of `1`? Or am I missing something?

```suggestion
        allow(client).to receive(:get_bib_info).with(catkey, 'fake-token').and_return(get_bib_info_response)
```

banukutlu: it is being used by the subject on top 

`subject(:builder) { described_class.new(catkey: catkey, user_token: 'fake-token', client: client, library: 'UP-PAT') }`

since get_bib_info is called after the prefix a is deleted, needed to stub with only `1`

cdmo: ah ok cool thanks

cdmo (approved):
Thanks!

PR #362, opened 2021-03-05
allows SymphonyClient to operate without the guests UserID and PIN

whereismyjetpack:


whereismyjetpack:
Instead of taking the guests userID from symphony, we get it from the REMOTE_USER variable. This will get set by Apache in qa/stage/prod. locally we can put this in with something like 

https://addons.mozilla.org/en-US/firefox/addon/modheader-firefox/

SymphonyClient was modified to use it's own credentials instead of ones supplied by the symphony callback. 

Instead of logging in and returning the current user, we login as "staff" and then lookup remote_user, and return that. 

This PR introduces two new configuration parameters
`Settings.symws.username` and `Settings.symws.password` We'll provide test accounts for developers to use. 

TODO

- [ ] test barred accounts
- [ ] test user logged in, but not in symphony (djb44)
- [ ] create test accounts for developer use
- [ ] configure non production servers with mod_openidc, and protect all paths
- [ ] any other testing that may need to take place.

banukutlu:
Just noting this here so to remember when testing:

image

```
Barred can't do anything
Delinquent can do anything
Blocked can renew and place holds only
Collection is the same as Blocked

Any of them should be able to update address
```

whereismyjetpack:
This code is running at https://myaccount-qa.dev.k8s.libraries.psu.edu (it's not being automatically deployed or anything, but i wanted an isolated place to test this)

whereismyjetpack:
Based off of our conversation in Teams yesterday, I coded up a crude "masquerade" feature. 

I Introduced a new `Settings` value of `admin_users` if the remote_user is in that list, and they initially login with the query parameter of `?masquerade=abc123` they will be switched to that user. If this is a useful feature, we could potentially expose it in the UI or something, but this gets Ruth unstuck from testing

Previously, when calling logout, it would redirect to the home page, now that the web-server is  enforcing authentication, it would just log you right back in. I coded up a crude logout view, so that the session doesn't get reinstated.

banukutlu:
> Just noting this here so to remember when testing:
> 
> image
> 
> ```
> Barred can't do anything
> Delinquent can do anything
> Blocked can renew and place holds only
> Collection is the same as Blocked
> 
> Any of them should be able to update address
> ```

@ruthtillman Have you been able to test with different accounts?

ruthtillman:
Yes, I have been able to test all of these!

.rubocop_todo.yml
```diff
@@ -9,7 +9,8 @@
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 258
+  # TODO change only for SymphonyClient, or refactor SymphonyClient
+  Max: 280
```
banukutlu, line 13: I think we can add an exclude for SymphonyClient to https://github.com/psu-libraries/myaccount/blob/main/.rubocop.yml

app/controllers/application_controller.rb
```diff
@@ -26,7 +26,10 @@ def patron
     MAX_INACTIVE_TIME = 2.hours - 5.minutes
 
     def authenticate_webaccess
-      redirect_to Settings.symws.webaccess_url + request.base_url
+      # if we aren't given a REMOTE_USER variable we are unauthorized.
+      # this maybe should be a 401, however if apache is misconfigured
+      # there's no amount of retrying that will fix this for the guest
+      redirect_to '/500' unless request.env.fetch(Settings.remote_user_header, nil)
```
banukutlu, line 32: we can redirect to a custom error page for this case as you mentioned @whereismyjetpack, we can add that in a later PR 👍  I think it is good for now, we all would love to see this PR deployed.

app/controllers/sessions_controller.rb
```diff
@@ -17,6 +17,10 @@ def index
       return redirect_to original_fullpath if original_fullpath.present?
 
       redirect_to summaries_url
+    else
+      # if the symphony client returns no user we might want to redirect to a page that
+      # says that they don't have a record?
+      redirect_to '/500'
```
banukutlu, line 23: we can redirect to a custom error page for this case as you mentioned @whereismyjetpack, we can add that in a later PR 👍 I think it is good for now, we all would love to see this PR deployed.

banukutlu (approved):
LGTM 👌  🎉   thanks @whereismyjetpack  🍻

PR #374, opened 2021-05-11
enable kubernetes

whereismyjetpack:
This PR introduces docker for local development, and starts the work for deploying to kubernetes. 

CI/CD follows the following pattern. 
main == qa
preview/foo == "dev" instance of "foo"
tags == production. 

infrastructure config lives at https://github.com/psu-libraries/myaccount-config this would take place of the myaccount bits of psulib_blacklight_deploy. 

This PR depends on #362 because i needed the OIDC bits in order to get a tested Kubernetes deployment 

- adds dockerfile, and docker-compose setup
- uses ERB template to pull configuration out of the environment for values that may be injected via ENV in docker (or through vault)
- adds some scripts in ./bin to aide in starting up or local development


whereismyjetpack:
@banukutlu -- I think this is ready to go. what should happen upon merging is https://myaccount-qa.dev.k8s.libraries.psu.edu should become available with this code. If there are any problems I can work to smooth them out. These changes _should_ only affect local development.

docker-compose.yaml
```diff
@@ -0,0 +1,77 @@
...
+      context: .
+      target: base
+    volumes:
+    - bundle-data:/app/vendor/bundle
+    - node-data:/app/node_modules
+    - type: bind
+      source: ./
+      target: /app/
+    ports:
+      - "3000:3000"
```
banukutlu, line 73: @whereismyjetpack would it be helpful to have this one configurable as well? we can just change it here but maybe makes more sense to do it in `.envrc` since it is not tracked?

whereismyjetpack: Yeah, we can do that. I think I'll also put examples in the .sample file

whereismyjetpack: I've added config bits for the app port, as well.

README.md
```diff
@@ -77,7 +77,76 @@ bin/webpack-dev-server
...
+```
+docker-compose down
+```
+
+Change the Redis, or App Port
+If you're running redis locally for another project, or you are running another rails project, you may need to change the port that's exposed for myaccount
+
+```
+echo "export REDIS_PORT=9999" >> .envrc
+echo "export APP_PORT=3001" >> .envrc
```
banukutlu, line 144: thanks for adding this too👌

.gitignore
```diff
@@ -3,9 +3,12 @@
 # If you find yourself ignoring temporary files generated by your text editor
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
+# Ignore secrets
+.envrc
 
 # Ignore bundler config.
 /.bundle
+.
```
banukutlu, line 12: @whereismyjetpack  do we need this one`/vendor/*` as well?

whereismyjetpack: I was going to address .gitignore as another PR since this one is getting super funk-a-tron. 

there's a few local paths. do you want me to add them in now?

banukutlu: it is totally fine with me to add more to gitignore, but yeah I hear you for when PRs gets biggers :)

whereismyjetpack: ok. I've added them to this one. I also added some docs on how rspec runs. (this process may continue to evolve. please provide feedback)

banukutlu (dismissed):
thanks @whereismyjetpack , I tried on my local, working nicely 👍

banukutlu (approved):

PR #383, opened 2021-06-04
Remove cdmo from list of default assignees

whereismyjetpack:


banukutlu (approved):
😭  for cdmo
🎉  for @whereismyjetpack

Issue #358, opened 2021-02-25
review more sirsi responses to exclude from datadog


banukutlu:
- [x] `Error 500: [{"code"=>"hatErrorResponse.17", "message"=>"User BARRED"}]`
- [x] [`Error 404: [{"code"=>"recordNotFound", "message"=>"Could not find a(n) /circulation/holdRecord record with the key null."}`](https://app.datadoghq.com/apm/trace/694420994088028051?spanID=1006731138935316895&env=production&sort=time&colorBy=service&spanViewType=metadata&graphType=flamegraph&shouldShowLegend=true) 
- [x] [`Error 500: [{"code"=>"acknowledgeRenewalFeeCircPrompt", "message"=>"Renewal fee of 22.00 USD acknowledgment required."}`](https://app.datadoghq.com/apm/trace/620730435647640378?spanID=3743199735654951147&env=production&sort=time&colorBy=service&spanViewType=errors&graphType=flamegraph&shouldShowLegend=true)
- [x] `Error 500: [{"code"=>"invalidPacket", "message"=>"Received incomplete packet: "}]`
- [x] [`Error 400: [{"code"=>"invalidDateFormat", "message"=>"The value '2021/06/31' is not in the appropriate date format."}`](https://app.datadoghq.com/apm/trace/1977042239201915520?spanID=3809736400005961838&env=production&sort=time&colorBy=service&spanViewType=metadata&graphType=flamegraph&shouldShowLegend=true)
- [x] [`Error 401: [{"code"=>"permissionDeniedRecordHidden", "message"=>"GET Permission denied: Record 3981350 hidden."}]`](https://app.datadoghq.com/apm/trace/694420994088028051?spanID=1006731138935316895&env=production&sort=time&colorBy=service&spanViewType=metadata&graphType=flamegraph&shouldShowLegend=true)
- [x] [`Error 500: [{"code"=>"hatErrorResponse.91", "message"=>"No more charges allowed"}]`](https://app.datadoghq.com/apm/trace/4163923067549482287?spanID=977628002360674633&env=production&sort=time&colorBy=service&spanViewType=metadata&graphType=flamegraph&shouldShowLegend=true)

PR #387, opened 2021-06-09
more sirsi response error codes to exclude from datadog

banukutlu:
#358 

These response codes are expected and does not need to be reported as error since we do not act on them.


whereismyjetpack:
🚀  I think this was your first `preview` deployment 🏅

PR #392, opened 2021-06-22
parameterizes more of docker-compose config

whereismyjetpack:
allows the user to change the SYMWS_* variables in their local .envrc file. SYMWS_URL no longer defaults so it must be set in the .envrc file


banukutlu (approved):
lgtm 👍🏻 danke!

PR #393, opened 2021-06-22
adds dependabot config for docker

whereismyjetpack:
our base image runs updates on security updates, when an update is found dependabot will issue a PR for update.


banukutlu (approved):
👍🏻 thanks!

Issue #396, opened 2021-06-28
Use high-res images for header and footer logos


nicktiberi:
Similar to the change for the catalog, let's use high-res images for My Account as well.

PR #397, opened 2021-06-28
Use high-res images for header and footer logos

nicktiberi:
Re: #396.

Similar to the recent change in the catalog, use the same high-res images and slight CSS tweaks in My Account to keep things visually consistent.

### Before:
Screen Shot 2021-06-28 at 2 19 59 PM

Screen Shot 2021-06-28 at 2 20 05 PM

### After:
Screen Shot 2021-06-28 at 2 20 12 PM

Screen Shot 2021-06-28 at 2 20 16 PM


banukutlu (approved):
👌🏻

PR #410, opened 2021-08-11
Pickup loc

ruthtillman:
Wasn't sure how the different tables were being used. This commit updates the part that displays after the hold is placed/in the holds management area.


banukutlu:
@ruthtillman  this branch needs a rebase, needed to pull the latest  but once merged we delete branch so not possible in this case so branch off of main again for the new changes is the way to go generally.

banukutlu:
now rebased

banukutlu (approved):

PR #411, opened 2021-08-11
rails upgrade

banukutlu:


awead (commented):
This all looks fine. Did you run `bundle exec rails app:update` and check the diffs? Another thing we might want to do is update the Rails-related npm packages. You can see which ones need updating with `yarn outdated`. This could also be a separate commit and/or PR.

banukutlu:
> @banukutlu This all looks fine. Did you run `bundle exec rails app:update` and check the diffs? There's probably nothing there since this is only a patch update. Another thing we might want to do is update the Rails-related npm packages. You can see which ones need updating with `yarn outdated`. This could also be a separate commit and/or PR.

yeah I ran rails app:update, which always shows bunch of diff due to frozen_string_literal, double quotes rather than single quotes etc. and some comments, there was one line I was not sure about in `config/environments/test.rb`, just checked my console output now and it is this line ` config.action_view.cache_template_loading = true`, I kinda thought that we might not want it true but what do you think?

I think that was it, if I did not miss any other changes among the other expected diff.

banukutlu:
> @banukutlu This all looks fine. Did you run `bundle exec rails app:update` and check the diffs? There's probably nothing there since this is only a patch update. Another thing we might want to do is update the Rails-related npm packages. You can see which ones need updating with `yarn outdated`. This could also be a separate commit and/or PR.

yeah I was thinking exactly to make yarn package updates as another PR, kinda want to get this deployed and see, last time rails upgrade with blackcat repo and ansible deployment did not play happy so wanted to keep the change minimal.

awead (approved):

PR #416, opened 2021-08-18
updates CI config

whereismyjetpack:
- created an org-global context in circle ci that has "defaulty" values for things like CI_IMAGE_UTILS_IMG_TAG, etc. we're updating the ci-utils image tag to work with the new version of harbor

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 jobs:
   release:
     docker:
-      - image: harbor.k8s.libraries.psu.edu/library/ci-utils:v1.0.6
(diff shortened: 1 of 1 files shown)


banukutlu:
@whereismyjetpack would you mind rebasing so to get the fix that gets the CI to hopefully pass :)

banukutlu (approved):
lgtm 👍🏻

whereismyjetpack:
still seems like a flake in some place, but it passed.

banukutlu:
> still seems like a flake in some place, but it passed.

yup def. still flakey :/ but yeah this PR looks good, no need to hold the PR up.

PR #417, opened 2021-08-18
allow datadog to be switched via ENV

whereismyjetpack:
in the config-repo we can set 

```
datadog:
  enabled: [true|false]
```

when switched in the helm chart we set the env appropriately (preview, or production) and then we also set the host to the cluster apm collector

I renamed `on_server` to `enabled` to follow the naming we use elsewhere. I can go either way on that change.


banukutlu (approved):
lgtm ✔️

PR #419, opened 2021-08-19
Json logging

whereismyjetpack:
- Removes the taggedlogger (since this info is in lograge output) 
- adds the json formatter so that logs in splunk are easier to read, parse, and ... 'splunk'

Here's a screenshot of the logs before, and after in splunk.

Before:
Screen Shot 2021-08-19 at 10 47 40 AM
After:
Screen Shot 2021-08-19 at 10 46 59 AM


config/environments/production.rb
```diff
@@ -70,7 +70,7 @@
   # config.active_support.deprecation = :notify
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
```
banukutlu, line 72: @whereismyjetpack  is this comment relevant?

lauterman (approved):

config/environments/production.rb
whereismyjetpack: it is not! mended.

banukutlu: thanks!

banukutlu (approved):
lgtm 👌🏻  thanks Dann, formatted logs are super helpful!

PR #424, opened 2021-08-20
latest updates to pickup locations list

banukutlu:
this is the other list which is used for translating sirsi code to human readable text on holds results table


whereismyjetpack (approved):

ruthtillman (approved):
looks good

PR #426, opened 2021-08-24
Catch errors when volumetric_natural_sort fails and log them

### LINKED ISSUE #404: Catkey 2972255 is failing
When trying to place a hold for catkey the following line is throwing an error

https://github.com/psu-libraries/myaccount/blob/main/app/services/place_hold_form/builder.rb#L102

`call.record['naturalized_volumetric'] = naturalize(call.volumetric.to_s)` is producing the following 

```
    "naturalized_volumetric"=>["v", 2.0]}>,
```

then when calling `.sort_by! it's failing because naturalized_volumetric is an array with 

```
ArgumentError: comparison of Array with Array failed
```

it works with calling sort_by on the first object, or if you call .to_s on it but i'm not sure what the desired outcome is, I feel like to sort by naturalized_volumetric it would need to be an integer, yes?

nicktiberi:
Closes #404.

myaccount attempts to sort volumetric data from Symphony, but sometimes there are issues with the data causing the sort to bomb. When this happens, we ignore the error and show the unsorted list, and also log an error.


banukutlu (approved):
lgtm ✔️  thanks @nicktiberi!

whereismyjetpack (approved):

PR #430, opened 2021-08-31
Update settings.yml

ruthtillman:
Altoona requested (and was approved by Lending Services) to repurpose UP-GATEWAY in the holds placement apparatus for their locker downtown. Edited the text in one list and added to the other just to be sure we have full coverage. It won't be people's default but they can choose it.


banukutlu (approved):
👍🏻

PR #432, opened 2021-09-14
log 401 from symphony

whereismyjetpack:
We've run into this particular error a few times in the past month, bubbles up a log message so we can more quickly remedy the problem.

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -21,6 +21,13 @@ class SymphonyClient
                          login: user_id,
                          password: password
                        })
+
+    if response.status > 400
+      Rails.logger.error(
+        "Received #{response.status} when attempting to login to #{Settings.symws.url}"
+      )
+    end
+
     resp = JSON.parse(response.body)
     session_token = resp['sessionToken']
     get_patron_record(remote_user, session_token)


app/services/symphony_client.rb
```diff
@@ -21,6 +21,13 @@ def login(user_id, password, remote_user = nil)
                          login: user_id,
                          password: password
                        })
+
+    if response.status > 400
+      Rails.logger.error(
+        "Received #{response.status} when attempting to login to #{Settings.symws.url}"
+      )
+    end
```
awead, line 29: Is there still something in the response body that allows the login process to continue? Does get_patron_record still return something?

whereismyjetpack: get_patron_record returns nil. but since the linter is barking at me, i'll apologize for wasting everyones time and just close it. since this seems to happen all the time, i'm stepping away, and will do work when explicitly asked to.

awead: There's definitely something to work on here. I'm ok with merging this and adding additional tickets to refactor. Sometimes pulling at the thread here and unravel into a lot more work. That doesn't mean we have to abandon it.

PR #437, opened 2021-09-27
null pickupLibrary fix

banukutlu:
when pickupLibrary is null in the response, holds page was failing 

it will display a default text in this case


banukutlu:
@nicktiberi @awead This is a hotfix that needs to be deployed, could one of you do a quick review please? Thanks!

nicktiberi (approved):
LGTM! 👍

Issue #423, opened 2021-08-20
Fix failing holds spec


banukutlu:
`rspec ./spec/features/holds_spec.rb:41 #`

https://app.circleci.com/pipelines/github/psu-libraries/myaccount/749/workflows/a838ef0f-0cd7-4eac-afe7-0410cb1eccf2/jobs/904

PR #445, opened 2021-11-04
Failing flakey hold spec to pending

banukutlu:
Changing this spec to pending until we have a fix for it #423 so the CI works for releases. 
We reviewed the behavior that is being tested with this spec on QA, it seems to be working OK.


nicktiberi (approved):
Good call. Someday we will fix this 🙏

banukutlu:
> Good call. Someday we will fix this 🙏

yup 🙏🏻  thanks!

PR #447, opened 2021-11-08
gets this var from circleci context

whereismyjetpack:


whereismyjetpack:
harbor changed the format for non-person accounts. this should fix it.

banukutlu (approved):

PR #477, opened 2022-03-17
adds symphony healthcheck

### LINKED ISSUE #472: Add Symphony WebServices to health checks

whereismyjetpack:
1. breaks out login, and get_patron record
2. returns status, and auth token from `login` this allows us to bubble up errors
3. refactors warden to use this method 
4. makes get_patron_record a public method, too
5. adds /health/symphony endpoint 

I'm not 100% sold on how this is done, but the idea is if there are errors in the application we can look at the logs, or look at /health/symphony or /health/all to see any error message that might bubble up 

Fixes #472

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,6 +22,7 @@ Metrics/ClassLength:
 Metrics/BlockLength:
   Exclude:
   - 'config/environments/production.rb'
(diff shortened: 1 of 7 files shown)


banukutlu (approved):
lgtm ✔️ 
thanks!

PR #519, opened 2022-08-11
preview/update ruby

whereismyjetpack:
- Fix circular argument reference
- updates to ruby 2.7.6
- updates redis
- refactor CI

CI now runs in the same env as docker. CI is slow because of pulls from harbor, I'm going to try to address that in another PR. 

Addresses #514


banukutlu (approved):
LGTM 👍🏻  thanks!

PR #546, opened 2022-11-28
No Recalls: ILLiad hold placement

### LINKED ISSUE #142: No Recalls: Create Illiad hold placement landing page
When users click through from the Catalog, if the hold should be placed through ILLiad, they land on a page which allows them to complete the hold. This page will include elements of the catalog record so that they don't have to fill these in. It will also contain a few optional elements that are ILL-specific (e.g. will they take an alternate or digital edition of the item)

Final contents of the page TBD with input from Access Services and ILL.

## WDDLL

- [x] Landing page uses webservices data fields including title, author, publication date, etc. 
- [x] Landing page allows user to make appropriate choices (not wanted after date, alternate editions of items, etc, from ILLIAD screen)
- [x] Landing page uses ILLiad APIs to actually place the hold.

## Possible Options to Discuss With Access Services

- allow user to choose to place a non-recalling hold instead -- this would be really easy to generate using the catkey. Appropriate text and then a link to the equivalent of: https://myaccount.libraries.psu.edu/holds/new?catkey=461483 

The conditions to send people here would be determined on the Blacklight side in https://github.com/psu-libraries/psulib_blacklight/issues/609.

### LINKED ISSUE #143: Send Request to Illiad
Rather than splitting the user's experience by redirecting them to Illiad, use the Illiad API to place the request: https://support.atlas-sys.com/hc/en-us/articles/360011809394-The-ILLiad-Web-Platform-API

This may involve figuring out methods to handle Illiad being slow or temporarily down (e.g. New School uses MySQL holding database and attempts to place hold every 5 min, updating status when hold placed).

What additional data should we collect in the hold form? What user account barriers might exist and how do we overcome them (perhaps 9.1 will solve this)?

banukutlu:
Closes #142
Closes #143


app/views/checkouts/ill_checkouts.html.erb
```diff
@@ -0,0 +1,33 @@
+<% if illiad_loans.present? %>
+  <table class="table table-responsive-sm">
+    <div class='row pl-3 pt-2 pb-3'>
+      <h1>Interlibrary Loan Checkouts</h1>
+      <p class='pl-2 pt-2'>
+        (<a href='https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll?Action=10&Form=10'>
```
whereismyjetpack, line 6: I wonder if this should be something like `Settings.illiad.manage_url` or something like this?

config/locales/en.yml
```diff
@@ -47,9 +47,9 @@ en:
       label_html: <i class="fa fa-exclamation-circle" aria-hidden="true"></i> Notice
       message: Returned materials may be quarantined for four days and will remain on your account during the quarantine period.
     checkouts_info_html: >
-      <p>Materials borrowed through <a href="https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll">Illiad</a> or <a href="http://www.libraries.psu.edu/mtss">MTSS</a> do not appear on this list and must be renewed separately.</p>
```
whereismyjetpack, line 50: Same with these, but i'm not sure if we can use ruby in locales?

whereismyjetpack (changes requested):

app/models/patron.rb
```diff
@@ -46,6 +46,18 @@ class Patron
     'HERSHEY'
   ].freeze
 
+  ILL_ELIGIBLE_PROFILES = [
```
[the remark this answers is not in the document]
banukutlu: yup thanks for the catch!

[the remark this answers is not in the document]
banukutlu: fixed 👍🏻

app/views/checkouts/ill_checkouts.html.erb
banukutlu: yup will do that

whereismyjetpack (dismissed):

ajkiessl (dismissed):

Smullz622 (approved):
LGTM!

PR #558, opened 2022-12-09
unbar link

### LINKED ISSUE #493: Why did my hold fail / user barred has linked to un-bar
Right now we have ways to unbar yourself from your My Account page and on the standalone page. When a user has a hold fail because they're barred AND they fit the criteria to accept lending policy, can we add a link to the error message so that they can fix it from that page vs. clicking to the My Account home page.

Screen Shot 2022-05-16 at 10 08 49 AM

Smullz622:
Display link to accept lending policy when patron's are barred because they have not yet accepted the policy.
Fixes #493 

Due to the way FakeSymphony is set up, I had to create a number of support files in order to create the 500 response. If there is a more efficient way around this, I'd be happy to learn & then refactor. Just let me know!

### DIFF
--- a/app/jobs/place_holds_job.rb
+++ b/app/jobs/place_holds_job.rb
@@ -30,7 +30,9 @@ class PlaceHoldsJob < ApplicationJob
 
       ApplicationController.render partial: 'holds/results',
                                    layout: false,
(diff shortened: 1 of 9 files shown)


banukutlu (approved):
lgtm 👍🏻  thanks @Smullz622!

banukutlu:
Merged to main with #569

PR #569, opened 2023-01-04
Accept lending policy rebase

### LINKED ISSUE #555: Better result checking for accept lending code request
Currently we are not checking if the request successful or not and redirecting to thank you page regardless.
We should check the response code and redirect accordingly here https://github.com/psu-libraries/myaccount/blob/main/app/controllers/lending_policy_controller.rb#L9

Smullz622:
closes #555


app/controllers/lending_policy_controller.rb
```diff
@@ -6,11 +6,16 @@ class LendingPolicyController < ApplicationController
   def accept
     if patron.eligible_for_wage_garnishment?
```
banukutlu, line 7:
```suggestion
      # even if the patron is not eligible for wage garnishment (already opted in or some other reason),
      # just redirect them to the thank you path to avoid further user confusion
      redirect_to lending_policy_thank_you_path unless patron.eligible_for_wage_garnishment?
      
```
 Rather than having nested if-else, here changing this one kinda like a guard clause reads a bit better imo, if we change it this way we won't need the else part and delete line 15-18.

spec/controllers/lending_policy_controller_spec.rb
```diff
@@ -46,15 +46,28 @@
       context 'when the patron is eligible for wage garnishment' do
```
banukutlu, line 46: Can we also add a spec for when patron is not eligible for wage garnishment?

banukutlu (changes requested):

app/controllers/lending_policy_controller.rb
Smullz622: thanks, it's fixed now

spec/controllers/lending_policy_controller_spec.rb
Smullz622: It's there, it just doesn't show up in the diff since it was there already

banukutlu (approved):
✔️  thanks @Smullz622

PR #576, opened 2023-01-13
Ruby to 3.1.3, Rails to 6.1.7

banukutlu:
Ruby  upgrade to 3.1.3, Rails upgrade to 6.1.7 and dependency updates

TODO:
- [ ] We had to downgrade Redis due to failing specs, need to look into it.
- [ ]  Need relevant CI changes, Docker file needs to be updated


whereismyjetpack:
I upgraded the Dockerfile, I had to downgrade datadog because of api changes to 1.0, and our MonkeyPatch will need reworked if we choose to continue using it. Aside from the one faling spec, I think this is good to go. 🎉

ajkiessl (approved):
This looks good to me.  I think the CI config needs updated still, though.

whereismyjetpack (approved):

PR #587, opened 2023-02-17
Alert added to settings.yml

ajkiessl:


whereismyjetpack:
This one can be done in myaccount-config, too

PR #601, opened 2023-03-28
'Recalled' and 'Overdue' checkouts

ajkiessl:
Adds 'Awaiting Recalled Processing' transaction status to checkouts query.  This displays as 'Recalled'.  Removes duplicate query from holds query.  Adds logic the displays for checkouts that displays 'Overdue' if we are past the due date.  Tests for all of this.


banukutlu (approved):
LGTM! 👍🏻

PR #605, opened 2023-03-30
add ILLiad materials to summary counts

### LINKED ISSUE #572: No-Recalls: Fix counts on My Account summaries page
Right now the checkouts and holds boxes on the homepage only reflect Symphony data. So I was seeing X checkouts when I have X + 1 because I have one ILL checkout. Similarly, wasn't seeing the holds box at all because my only hold was in ILLiad.

The status maps to checked out and holds numbers/sub-numbers we display would be:

## Checked Out

- checked out: "Checked Out to Customer" & starts with "Renewed by"
- overdue: as above but due date in past?
- recalled: "Awaiting Recalled Processing"

## Holds

- on hold: anything that goes into holds except Customer Notified via E-Mail
- available: Customer Notified via E-Mail

Note: Talk to Ruth for statuses for ready to pick up, picked up etc.

Smullz622:
Fixes #572 

Also fixes ILL hold bug: With Ruby 3 upgrade, positional vs keyword arguments were separated. This caused some errors with the SymphonyClientParser that were resolved.

### DIFF
--- a/app/services/symphony_client_parser.rb
+++ b/app/services/symphony_client_parser.rb
@@ -2,9 +2,8 @@
 
 # Responsible for parsing SymphonyClient responses.
 class SymphonyClientParser
-  def self.parsed_response(symphony_client, symphony_call, *params)
-    kwargs = params.first.nil? ? {} : params.first
-    client_response = symphony_client.send(symphony_call, **kwargs)
-    JSON.parse client_response.body
+  def self.parsed_response(symphony_client, symphony_call, **params)
+    client_response = symphony_client.send(symphony_call, **params)
+    JSON.parse(client_response.body)
   end
 end
--- a/app/services/symphony_client_parser.rb
+++ b/app/services/symphony_client_parser.rb
@@ -2,8 +2,9 @@
 
 # Responsible for parsing SymphonyClient responses.
 class SymphonyClientParser
-  def self.parsed_response(symphony_client, symphony_call, **params)
-    client_response = symphony_client.send(symphony_call, **params)
+  def self.parsed_response(symphony_client, symphony_call, *params)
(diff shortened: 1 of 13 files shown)


app/services/symphony_client_parser.rb
```diff
@@ -2,8 +2,9 @@
 
 # Responsible for parsing SymphonyClient responses.
 class SymphonyClientParser
-  def self.parsed_response(symphony_client, symphony_call, **params)
-    client_response = symphony_client.send(symphony_call, **params)
+  def self.parsed_response(symphony_client, symphony_call, *params)
+    kwargs = params.first.nil? ? {} : params.first
```
whereismyjetpack, line 6: Should we just pass `params` instead of `params.first` ? Are there other places where we're calling `symphony_client.send` that may need refactoring, too?

Smullz622: `params` gives a hash inside of an array, so first gives us just the hash that we need

We don't call `symphony_client.send` anywhere else, and it looks like the other calls for the parser were already reformatted.

banukutlu: When I try with **params, 
```
def self.parsed_response(symphony_client, symphony_call, **params)
  client_response = symphony_client.send(symphony_call, **params)
  JSON.parse(client_response.body)
end
```
ILL holds working on my local. Do we need this change or I might be missing sth here?  I think the change in the bib_info method fixes it?

ajkiessl: Yeah I think the change in the bib_info is probably sufficient if *params is changed back to **params.  Might still have to handle nils with the **kwargs part, though.

banukutlu (approved):
LGTM! thanks Sarah! 🚀

PR #608, opened 2023-03-31
Creates ILLiadUser

whereismyjetpack:
- adds create_user, and user_exists? methods in IlliadClient
- adds email_address to Patron model 
- pulls address1 from SymphonyClient (this is where email is stored)
- introduces a symphony library -> Illiad Site mapping in settings 

TODO

- [ ] add all the symphony to illiad site mappings
- [x] test create_user

Questions [& RKT updates to questions]
What should the default library be, if we don't have a mapping? [RKT answer: Pattee for statistical likelihood of it being correct.]

We are not able to easily test this code, because we can not delete users out of the illiad platform. We might want to investigate a better way of testing this integration, I can see the lack of a test internment really biting us in the future -- especially when illiad decides to make platform changes. [RKT answer: talking to Shane about setting y'all up with the below reference.]


whereismyjetpack:
If I could get access to this thing, it might be a help 
https://help.oclc.org/Resource_Sharing/ILLiad/Troubleshooting/How_to_delete_an_ILLiad_Patron_account
@ruthtillman

ruthtillman:
@whereismyjetpack I'll talk with Shane about the deletion thing! It might be feasible but require Windows emulation or somesuch :-/

PR #614, opened 2023-04-12
suggest local library for ILL ineligible

### LINKED ISSUE #578: No Recalls: Suggest PA residents try their local libraries
When we have a PA Borrower who attempts to get an item that would be ILLed, we'll direct them to the normal hold page to place a non-recalling hold. Suggestion from Access Services Council, could we embed something in the page (something conditional I guess?) suggesting that they might be able try their local library? Message would be something like:

`A local copy isn't available. You can join the holds list for when one is returned. Also consider checking at your local library or asking if they can get a copy for you through Interlibrary Loan.`

(placeholder message for if we do actually get it done)

This is a nice-to-have. If it's smooth to fit into MVP while setting up behaviors/account creation, then that's cool. Otherwise we'll add it to later work on My Account.

Smullz622:
Fixes #578 

I spoke with Ruth & we're extending this to any patrons who are not eligible for ILL, not just PA residents.
Patrons who are not ILL eligible will see the following message:

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -287,8 +287,6 @@ class SymphonyClient
         ["circRecordList{*,#{ITEM_RESOURCES}}"]
       when ->(h) { h[:holdRecordList] }
         ["holdRecordList{*,#{ITEM_RESOURCES}}"]
-      when ->(h) { h[:address1] }
-        ['address1']
       when ->(h) { h[:all] }
         [
           'blockList{*}',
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
LGTM

app/services/symphony_client.rb
```diff
@@ -287,8 +287,6 @@ def patron_linked_resources_fields(item_details = {})
         ["circRecordList{*,#{ITEM_RESOURCES}}"]
       when ->(h) { h[:holdRecordList] }
         ["holdRecordList{*,#{ITEM_RESOURCES}}"]
-      when ->(h) { h[:address1] }
```
banukutlu, line 290: thanks for cleaning this out 👍🏻

banukutlu (approved):
LGTM!

PR #618, opened 2023-04-18
sends call number instead of barcode for ILL volumetric requests

### LINKED ISSUE #613: For ILL volumetric, is it possible to send call # vs. item ID?
ILL staff _can_ use the itemID/barcode to look up which volume is intended, but volume/part info is actually in the call number and it would be easier for them if that's what we send to ILLiad.

We're displaying the call number when letting people pick volumes, so I'm hoping that the answer is yes! 🤞 Whatever the value of the checkbox is, just send that.

Smullz622:
Fixes #613 
Also fixes flaky features/holds spec


banukutlu (approved):
LGTM 👍🏻

PR #619, opened 2023-04-18
Adds 'Pattee Notices' to holds query

### LINKED ISSUE #616: Adding "Pattee Notices" to Holds page Processing Statuses
From Shane:

> I discovered a status we missed for our ILLiad items.  We have a temporary queue that all Pattee pickup items go into while we get them ready (put paper bands on, etc.) to take down to circulation. 
> 
> "Pattee Notices"
>
> A request in this status is currently disappearing from the my accounts display.  Not a big deal but if you could add it to the statuses that say in process, that would be great.  I don't think there will be too many patrons checking their account during the 30-60 minutes the request is in this status but when you get around to it, it'd probably help.

ajkiessl:
closes #616

### DIFF
--- a/app/services/illiad_client.rb
+++ b/app/services/illiad_client.rb
@@ -96,6 +96,7 @@ class IlliadClient
         'Cancelled by Customer',
         'Duplicate Request Review',
         'Request Available Locally',
+        'Pattee Notices',
         'LST TESTING'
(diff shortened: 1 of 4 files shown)


banukutlu (approved):
LGTM!

Issue #565, opened 2022-12-16
No-Recalls: Create Illiad user


whereismyjetpack:
When placing a loan on behalf of a user, if the user does not exist in illiad we need to create the user. 

In looking at this we might need to get more information out of symphony for the user. 

We'll need to match up symphony library with illiad "site" I think 

Below are the following fields with what I think they should be matched up with 

| Illiad                   | Patron                                                                  |
| ----------------| ---------------------------------------------- |
| UserName          | patron.id                                                              |
| ExternalUserId   | patron.id                                                              |
| LastName          | patron.last_name                                                 |
| FirstName          | patron.first_name                                                |
| EmailAddress    | presently not available, address1 in symphony   |
| Site                    | patron.library                                                        | 

(RKT:) Other default fields/values for creation:

| Field | Value |
| --- | --- |
| DeliveryMethod | "Hold for Pickup" |
| LoanDeliveryMethod | "Hold for Pickup" |
| NotificationMethod | "Electronic" |

```
  "UserName": "", 
  "ExternalUserId": "",
  "LastName": "",
  "FirstName": "",
  "EmailAddress": " 
  "Phone": null,
  "Department": "Libraries",
  "NVTGC": "UPILL",
  "NotificationMethod": "Electronic",
  "DeliveryMethod": "Hold for Pickup",
  "LoanDeliveryMethod": "Hold for Pickup",
  "LastChangedDate": null,
  "AuthorizedUsers": null,
  "Cleared": "Yes",
  "Web": true,
  "Address": null,
  "Address2": null,
  "City": null,
  "State": null,
  "Zip": null,
  "Site": "Pattee Commons Services Desk",
  "ExpirationDate": "2024-11-22T00:00:00",
  "Number": null,
  "UserRequestLimit": null,
  "Organization": "University Park",
  "Fax": null,
  "ShippingAcctNo": null,
  "ArticleBillingCategory": null,
  "LoanBillingCategory": null,
  "Country": null,
  "SAddress": null,
  "SAddress2": null,
  "SCity": null,
  "SState": null,
  "SZip": "article",
  "SCountry": null,
  "RSSID": "",
  "AuthType": "Default",
  "UserInfo1": null,
  "UserInfo2": null,
  "UserInfo3": null,
  "UserInfo4": null,
  "UserInfo5": null,
  "MobilePhone": null
}
```

Site to Library Mapping:

| Library                   | Site                                         |
| ----------------| ------------------------------- |
[...]

PR #621, opened 2023-04-19
Create Illiad Account

### LINKED ISSUE #617: Checking ILLiad User Status
We're already handling two possible cases: if they already have an account and if they need one created.

However, there's an occasional third case that my arise where a person doesn't fall into resident/etc cases but is also not ILLiad eligible. All users will be cleared by folks in ILL and they will block those who shouldn't get an account (and send them a nice note/help them with Symphony hold).

On our application side, we need to be sure that we detect cases when a person:
1. has an account
2. that account is blocked or disavowed

and just route them to regular hold placement as well. We'll need to see what the API returns.

Data from Shane:

```
jsa5227 is currently blocked if you want to try it to see what happens.

tae5199 is disavowed
```

ILLiad made a list of statuses:

Allowed Values | Description
-- | --
Yes | The user is allowed to use the system.
No | The user has not yet been vetted by staff.
New | User is new to the system and has not completed  their user profile. On their first login they will be forced to  complete their user profile.
DIS | The user is disavowed and not authorized to log in.
B | The user is blocked.
BO | The user is blocked due to an overdue. Overdue blocks are removed when the user no longer has overdue materials checked out.
BX | The user is blocked by an external system.

whereismyjetpack:
Fixes #608 and #620 

Closes #617 and #565 

TODO 

- [x] finalize mapping of symphony location to illiad site.


whereismyjetpack:
I updated the locations, and I think this is finally ready to go.

banukutlu (approved):
LGTM 🚀

PR #630, opened 2023-04-27
updates illiad urls

### LINKED ISSUE #627: Update some ILLiad URLs for auth change
On May 10th at 8am, they'll be trying to switch over auth again. This means that we should be prepared to deploy about 9am.

We'll want to replace any instances with a base URL of: `https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/`

with

`https://psu.illiad.oclc.org/` which we use in a number of places already.

I believe this is just in config/locales/en.yml and config/settings.yml

We may want to set this one up ahead of time and test to see if the new URL just works. If so, we can deploy right away. I think it should for the ways it's used here.

Smullz622:
Fixes #627 

It seems like the new links are already working, so this can be merged/deployed

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -80,5 +80,5 @@ en:
         accept_ebook: If the item is found to be available digitally within Penn State's subscriptions, will you use the e-book?
       place_loan:
         success_html: For interlibrary loan requests, you will receive a notice once items are available for pick up.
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
👍🏻

ajkiessl (changes requested):
This will require rebasing with main.  I moved the "manage_url" out of settings.yml and into the en.yml, so that will need to be updated.

ajkiessl (approved):

PR #631, opened 2023-05-01
Integrate Hershey Users

### LINKED ISSUE #622: Sending Hershey folks to Hershey ILLiad
We've worked out with Hershey that their statuses and policies for this process will be exactly the same as the ones we're currently using for the rest-of-PSU-ILLiad. 

Hershey's ILLiad will be the same params, etc., except they will be `mhy` instead of `upm`.

We would have to determine the most efficient way to direct users there. They should all have the home library of HERSHEY in Symphony. So one way might be to split them out at the point at which we're splitting off the folks who don't get ILL at all and then set a parameter of whether it should be UPM or MHY.

ajkiessl:
Grabbing holds and checkouts from Illiad works the same for Hershey users.  Creating loans works the same, too.  The "manage ill" links needed to change depending on whether or not the user is from Hershey.  Creating Hershey users requires some of the metadata sent to Illiad to be different.

closes #622

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -81,4 +81,5 @@ en:
       place_loan:
         success_html: For interlibrary loan requests, you will receive a notice once items are available for pick up.
         error_html: <strong>Sorry!</strong> there was an error in requesting this item. Please <a href="https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/%{library}/illiad.dll">request it directly through our Interlibrary Loan form</a> or contact our ILL department at <a href="mailto:UL-ILLIAD-SUPPORT@psu.edu">UL-ILLIAD-SUPPORT@psu.edu</a>.
+      manage_url: https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/%{library}/illiad.dll?Action=10&Form=10
     loading_html: Items are loading. Longer lists of items will take a few moments to load.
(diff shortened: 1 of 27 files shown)


config/locales/en.yml
```diff
@@ -80,5 +80,6 @@ en:
         accept_ebook: If the item is found to be available digitally within Penn State's subscriptions, will you use the e-book?
       place_loan:
         success_html: For interlibrary loan requests, you will receive a notice once items are available for pick up.
-        error_html: <strong>Sorry!</strong> there was an error in requesting this item. Please <a href="https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll">request it directly through our Interlibrary Loan form</a> or contact our ILL department at <a href="mailto:UL-ILLIAD-SUPPORT@psu.edu">UL-ILLIAD-SUPPORT@psu.edu</a>.
+        error_html: <strong>Sorry!</strong> there was an error in requesting this item. Please <a href="https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/%{library}/illiad.dll">request it directly through our Interlibrary Loan form</a> or contact our ILL department at <a href="mailto:UL-ILLIAD-SUPPORT@psu.edu">UL-ILLIAD-SUPPORT@psu.edu</a>.
+      manage_url: https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/%{library}/illiad.dll?Action=10&Form=10
```
Smullz622, line 84: FYI this url base is changing May 10th (see #627). The current PR changes it for `error_html`, but we'll need to update it for `manage_url` as well.

Smullz622 (approved):
LGTM

PR #633, opened 2023-05-08
Symphony custom user not found page

### LINKED ISSUE #375: Add an error page for when symphony client returns no user
https://github.com/psu-libraries/myaccount/pull/362#discussion_r631959466

Some of the work done with #477 is relevant to this issue.

ajkiessl:
I made the status code for this 500, which is what is was before.  Not sure if 401 would be better, but I did try to make it a 401 and for some reason rspec can't handle that.

closes #375


Smullz622 (approved):
lgtm

whereismyjetpack (approved):

PR #642, opened 2023-06-15
updated matomo script

### LINKED ISSUE #639: Matomo: Add tracking code to My Account
Add it here so that we can start tracking some high-level stuff like ill holds vs. catkey holds.

Smullz622:
Fixes #639 

Turns out the Matomo script was already there, but the id_site was pointing to stage. This should point it to prod.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -29,7 +29,7 @@ redis:
     uri: <%= ENV.fetch("REDIS_SIDEKIQ_URI") { "redis://127.0.0.1:6379/1" } %>
   database:
     uri: <%= ENV.fetch("REDIS_DATABASE_URI") { "redis://127.0.0.1:6379/2" } %>
-matomo_id: 13
+matomo_id: 38
 illiad_locations:
   ABINGTON: Abington
   ALTOONA: Altoona


app/views/layouts/_head.html.erb
```diff
@@ -13,22 +13,21 @@
...
+  <!-- Matomo -->
+  <script>
+    var _paq = window._paq = window._paq || [];
+    /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+    _paq.push(['trackPageView']);
+    _paq.push(['enableLinkTracking']);
+    (function() {
+      var u="https://analytics.libraries.psu.edu/matomo/";
+      _paq.push(['setTrackerUrl', u+'matomo.php']);
+      _paq.push(['setSiteId', '13']);
```
whereismyjetpack, line 25: we shouldn't hardcode the site_id here. I think we were doing it right before, with `Settings.matomo_id` I think we just need to update the matomo_id in the prod deployment to be 11.

whereismyjetpack (changes requested):

whereismyjetpack:
I've created a non-production myaccount website in matomo, it's site_id is 38. You can add this into settings, and then in prod we have overridden in to be 14

Smullz622:
> I've created a non-production myaccount website in matomo, it's site_id is 38. You can add this into settings, and then in prod we have overridden in to be 14

Should the override be in `config/settings/production.yml` or in the helm chart?

config/settings.yml
```diff
@@ -29,7 +29,7 @@ redis:
     uri: <%= ENV.fetch("REDIS_SIDEKIQ_URI") { "redis://127.0.0.1:6379/1" } %>
   database:
     uri: <%= ENV.fetch("REDIS_DATABASE_URI") { "redis://127.0.0.1:6379/2" } %>
-matomo_id: 11
+matomo_id: 13
```
whereismyjetpack, line 32: @Smullz622 I would change this to be 38, and we'll leave the override in the config-repo!

whereismyjetpack (approved):

PR #654, opened 2023-08-29
removes UP-ARCHIT

### LINKED ISSUE #653: Temporarily remove Architecture & Landscape Architecture as pickup location in My Account
This is high priority but small -- because Architecture, UP-ARCHIT will be closed for an unknown amount of time, Access Services would like it removed ASAP from pickup locations. We'll restore it when it's open again.

It's in config/settings.yml and in https://github.com/psu-libraries/myaccount/blob/main/spec/support/data/other/hold_lookup_3912343.json 

(With closures like York, we still bring holds to the campus and they're picked up elsewhere. But because this is a branch library, people just have to choose another UP site.)

whereismyjetpack:
Fixes #653


ruthtillman (dismissed):
Looks good to me, thank you.

whereismyjetpack:
@ajkiessl I had to fix CI also to unblock this one.

ajkiessl (approved):

PR #665, opened 2023-10-23
updates sidekiq, and simplecov, redis

whereismyjetpack:
Updates sidekiq. Simplecov needed updating at the same time, redis was also updated at this time


ajkiessl (approved):
Looks good.  I guess we can update simplecov for other things now.  It was locked at < v0.18.0 for a while.

PR #677, opened 2023-12-13
export checkouts as RIS file

### LINKED ISSUE #675: Investigate: Feasibility of Exporting Current checkouts as RIS
Because we are about to get rid of checkout histories, public services is interested in mitigation strategies that help folks keep track of what they've read (but make it their choices).

Is it possible to repurpose the export all bookmarks as RIS code work we just finished to allow folks to export all current checkouts as an RIS file? (A bonus might be doing the same with emailing yourself a copy of it, but this isn't as high priority)

Whether it's adding within My Account or a button that sends the person's catkeys over to the Catalog to handle the function, it would be located in the Checkouts/Renewals page.

== 

Email format for Catalog records from Symphony:

```
catalog URL: bib => key (appended to URL)
Title: bib => fields => title
Author: bib => fields => author
Call #: call => fields> dispCallNumber
```

Email format for ILL, based on fields from ILLiad:

```
Title: LoanTitle
Author: LoanAuthor
Date: LoanDate
Identifier: ISSN (why god!)
```

Smullz622:
fixes #675 

This adds export buttons to the symphony and ILL checkouts tables to either download all respective checkouts as an RIS file or send a list of checkouts as an email.

Symphony RIS download- It pulls the catkeys for each checkout & sends them to catalog. The actual download process is happening in the catalog (https://github.com/psu-libraries/psulib_blacklight/pull/1256), but with how routing is structured there, users will appear to never leave MyAccount. It had to be separated this way because only MyAccount has the checkouts catkeys and only catalog has the item metadata needed for a decent RIS.

ILL RIS download- While some ILL records might have catkeys in `ItemInfo2`, they don't always. Since we can't consistently get them, we're just doing a separate RIS download process for ILL checkouts. It doesn't have as many of the fields for the RIS file, but the decision was that something is better than nothing.


ajkiessl (approved):
LGTM.  It might be a good idea to add a test in catalog that tests the parameters coming in from the path.

ajkiessl:
Ah, I didn't realize that MyAccount never had any mailers set up (even had actionmailer disabled).  Since mailing is now a part of this application I think it would be a good idea to work with @whereismyjetpack to get mailcatcher set up in k8s.  Then maybe deploy a preview branch to make sure emails are being sent.

ajkiessl:
I also think the `config/environments/production.rb` will have to be configured similarly to catalog's:

```
  config.action_mailer.delivery_method = :smtp

  config.action_mailer.smtp_settings = {
    address: ENV.fetch('SMTP_ADDRESS', 'localhost'),
    port: ENV.fetch('SMTP_PORT', 25),
    user_name: ENV.fetch('SMTP_USERNAME', nil),
    password: ENV.fetch('SMTP_PASSWORD', nil),
    authentication: ENV.fetch('SMTP_AUTHENTICATION_TYPE', nil)
  }
  ```
  
So we'll need those env variables set in the pods.

File: app/views/checkouts/_checkouts_table.html.erb [hunk 1]
```diff
@@ -1,6 +1,29 @@
...
     <table id="checkouts-table" class="table table-responsive-sm" role="table" aria-live="polite">
-      <caption><h2>Checkouts</h2></caption>
+      <caption>
+        <span class="dropdown float-right">
+          <style>
+            .dropdown button,
+            .dropdown-menu {
+              font-family: "Open Sans", Arial, serif;
+            }
+          </style>
```
Smullz622, line 10: @ajkiessl I was able to fix the font by adding the styling here. I tried putting it in a few different stylesheets instead, but it didn't seem to work anywhere else.  I don't love it being in the html though so I'm open to suggestions

ajkiessl: You could maybe try adding a separate class to a stylesheet like `.dropdown-btn-font` and use the `!important` property with it.  So maybe something like this:

```
.dropdown-btn-font {
  font-family: "Open Sans", Arial, serif !important;
}
```

Then add that class to the dropdown menu or button.

app/views/checkouts/_checkouts_table.html.erb
```diff
@@ -1,6 +1,29 @@
...
+            id="export-checkouts-btn"
+            data-toggle="dropdown"
+            aria-haspopup="true"
+            aria-expanded="false">
+            Export All Checkouts
+          </button>
+          <div class="dropdown-menu" aria-labelledby="export-checkout-btn">
+            <p><%= link_to 'RIS File', "#{Settings.catalog.bulk_ris_url}#{(checkouts.map { |c| c.catkey.to_s }).join(',')}" %></p>
+            <%= link_to 'Email', export_checkouts_email_path(checkouts: checkouts.map { |c| { catkey: c.catkey, title: c.title, author: c.author, call_number: c.call_number } }) %>
+          </div>
```
ajkiessl, line 23: To format this dropdown in the bootstrap way i think we want it to look like this:

```
<div class="dropdown-menu" aria-labelledby="export-checkout-btn">
      <%= link_to 'RIS File', "#{Settings.catalog.bulk_ris_url}#{(checkouts.map { |c| c.catkey.to_s }).join(',')}", class: 'dropdown-item' %>
      <%= link_to 'Email', export_checkouts_email_path(checkouts: checkouts.map { |c| { catkey: c.catkey, title: c.title, author: c.author, call_number: c.call_number } }), class: 'dropdown-item' %>
</div>
```

Reference: https://getbootstrap.com/docs/4.0/components/dropdowns/

app/views/checkouts/ill_checkouts.html.erb
```diff
@@ -5,6 +5,21 @@
...
+          id="export-ill-checkouts-btn"
+          data-toggle="dropdown"
+          aria-haspopup="true"
+          aria-expanded="false">
+          Export All ILLiad Checkouts
+        </button>
+        <div class="dropdown-menu" aria-labelledby="export-ill-checkout-btn">
+          <p><%= link_to 'RIS File', export_ill_ris_path %></p>
+          <%= link_to 'Email', export_ill_checkouts_email_path %>
+        </div>
```
ajkiessl, line 21: Another bootstrap dropdown change:

```
<div class="dropdown-menu" aria-labelledby="export-ill-checkout-btn">
   <%= link_to 'RIS File', export_ill_ris_path, class: 'dropdown-item' %>
   <%= link_to 'Email', export_ill_checkouts_email_path, class: 'dropdown-item' %>
</div>
```

ajkiessl (changes requested):
I have a couple change requests.  Everything else looks good, though.

ajkiessl (approved):
LGTM!

PR #681, opened 2024-01-05
email fix

Smullz622:
removes checkout data as a param & instead makes a symphony call to get the data. This cleans up the url & fixes the issue where large numbers of checkout exports (70+) were failing to process

fixes ILL RIS bug- multiple checkouts were being sent as one large item because the new line wasn't registering in single quotes


ajkiessl (approved):
Looks good!

PR #684, opened 2024-01-08
resolves CSRF vulnerability

Smullz622:
skipping token verification created a cross site request forgery vulnerability

This removes the line that skipped verification

### DIFF
--- a/app/controllers/redis_jobs_controller.rb
+++ b/app/controllers/redis_jobs_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class RedisJobsController < ApplicationController
-  skip_before_action :verify_authenticity_token
-
   def show
     job_status = Redis.current.get(params[:id]) || { result: :not_found }
     render json: job_status


whereismyjetpack:
we should deploy this and verify that background jobs are working before merging.

Smullz622:
> we should deploy this and verify that background jobs are working before merging.

@whereismyjetpack by deploy, do you mean switch it to a preview branch?

ajkiessl:
I agree with Dann.  We should at least confirm that the polling (`app/javascript/submission_handling/polling.js`) works with CSRF protection turned back on.  We might even want to add a test for it.

Smullz622:
Preview is up https://myaccount-csrf-vuln.dev.k8s.libraries.psu.edu/

It seems like everything is working to me. I was able to place an ILL hold, place/cancel a regular hold, and all the tables were able to correctly retrieve checkout/hold data.

whereismyjetpack (approved):

PR #702, opened 2024-04-05
Rails 7 Upgrade

ajkiessl:
Upgrades rails 6 to rails 7.  Update listen and rspec-rails so they work with rails 7. Fixes deprecations for rails 7.


whereismyjetpack:
we should put this out on a preview to make sure puma still works properly, but lgtm

ajkiessl:
new PR for preview: #703

PR #724, opened 2024-07-15
add engineering library back

Smullz622:
add engineering library back with new address upon reopening

do not deploy until 8/26


jlandiseigsti (approved):
👍

PR #728, opened 2024-07-19
handle timeouts in illiad_client

whereismyjetpack:


whereismyjetpack:
We refactored the illiad client to always return something so that a raised exception doesn't crash the summaries page. 

we will be writing up a ticket to, instead, refactor the summaries page in a future cycle

Smullz622 (commented):
Changes look good. Since we are planning to refactor anyway, we can just xit `./spec/services/illiad_client_spec.rb:216` and `./spec/services/illiad_client_spec.rb:287` for now. They're failing because we're now rescuing the error.

whereismyjetpack:
@ajkiessl @jlandiseigsti -- i'd like your review too before releasing. thanks!

app/jobs/view_illiad_loans_job.rb
```diff
@@ -12,16 +12,22 @@ def perform(webaccess_id:, type:, library_key:)
...
-    Redis.current.set("view_ill_#{type}_#{webaccess_id}", {
-      result: :success,
-      html:
-    }.to_json)
-    nil
-  rescue RuntimeError => e
-    process_failure(error_message: e.message, webaccess_id:, type:)
+    # IlliadClient usually returns an array, If we don't return an array we send the
+    # result to process_failure. We do this to avoid raising an exception in IlliadClient
+    if illiad_loans.instance_of?(Array)
```
jlandiseigsti, line 17: I take it that there are no "non-Array but still legit" results we might get?

jlandiseigsti (approved):
Seems good so far.

app/jobs/view_illiad_loans_job.rb
whereismyjetpack: We should always be returning an array in the client for successful calls, even if there's one item.

ajkiessl (approved):
lgtm

PR #843, opened 2026-07-29
Rails 8 upgrade

### LINKED ISSUE #839: Upgrade to Rails 8.1

Smullz622:
Fixes #839 

It seems like part of what was creating problems is that rails ujs is still being implemented. I put some quick fixes here and created a ticket to remove references and replace with turbo as needed (#847 ). I don't want to hold up the rails upgrade since that is already depreciated so we'll get to it when we work on catalog shortly.


config/environments/production.rb
```diff
@@ -1,121 +1,82 @@
...
-  params_to_include = %w(controller action)
-  config.lograge.custom_options = lambda do |event|
-    params = event.payload[:params].reject { |k| params_to_include.include?(k) }
-    {
-      params:,
-      time: Time.now
-    }
-  end
-  hostname = Socket.gethostname || 'production'
-  config.paths['log'] = "log/#{hostname}.log"
```
ajkiessl, line 96: Removing this might affect the log output in production.  I left the lograge stuff in the rails 8 upgrades I did, just to keep things the same.  Iirc it was to output the logs in JSON so they could be parsed by splunk more easily.  I'm not sure if that still applies here.

ajkiessl (dismissed):
I meant to approve with my comment above.  Everything else looks good.  Changing the logging output might not affect much, but I haven't tested out how that might change in production.

config/environments/production.rb
Smullz622: I'm not sure either. I put it back to be safe

ajkiessl (approved):
LGTM

